Issue #9, opened 2023-09-01
Update FENGSHA dust and other dust schemes


bbakernoaa:
The fengsha dust scheme is fairly old here and is based on the original GEFS-Aerosol implementation.  It needs updates.

Also the dust schemes are currently hard coded to use 5 dust bins.  We need to make this arbitrary with the schemes.

bbakernoaa:
This is being moved to https://github.com/ufs-community/CECE/pull/6

PR #10, opened 2023-09-01
Update FENGSHA and dust schemes

bbakernoaa:
This PR address #9 

Major changes are: 

- Update to all dust schemes to allow an arbitrary number of bins in each of the schemes
- Code cleanup for each of the dust schemes 
- Update fengsha to the latest version used in the UFS, i.e. RRFS-SD, AQM, and UFS-Aerosols


parameters/dust_data_mod.F90
```diff
@@ -89,7 +58,7 @@ module dust_data_mod
     (/      1.,     0.2,     0.2,     0.2,     0.2,     0.2,   0.333,   0.333,   0.333 /)
 
 
-  ! -- soil vagatation parameters
+  ! -- soil vegatation parameters
```
zmoon, line 61:
```suggestion
  ! -- soil vegetation parameters
```

bbakernoaa: fixed

src/dust_afwa_mod.F90
```diff
@@ -7,9 +7,7 @@ module dust_afwa_mod
 !  06/2023, Restructure for CATChem, Jian.He@noaa.gov
 
   use catchem_constants, only : kind_chem, g=>con_g, pi=>con_pi
-  use catchem_config, only: num_chem,num_emis_dust,&
-                            p_dust_1,p_dust_2,p_dust_3,p_dust_4,p_dust_5, &
-                            p_edust1,p_edust2,p_edust3,p_edust4,p_edust5
+  use catchem_config, only: num_chem,num_emis_dust,p_dust_1, p_edust1
```
jianheACM, line 10: This may be related to another issue. I'm thinking of undefining num_chem in catchem_config.  num_chem would be different for different configurations. Maybe we can assign ntchm to num_chem?

src/gocart_dust_default_mod.F90
```diff
@@ -2,209 +2,225 @@
...
-
-    ! -- for output diagnostics of dust source 
-    srce_dust(p_edust1)=srce_out(1)
-    srce_dust(p_edust2)=srce_out(2)
-    srce_dust(p_edust3)=srce_out(3)
-    srce_dust(p_edust4)=srce_out(4)
-    srce_dust(p_edust5)=srce_out(5)
+    call source_du( ndust, dt, tc, erodin, ilwi, area, w10m, gwet, rho_phy, airmas, bems,srce_out, start_month, g, ipr)
+
+    do n = 0, ndust-1
```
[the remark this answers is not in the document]
bbakernoaa: Yes thats what this loop does. The assumption is that ndust in gocart is the number of dust bins set in the parameters and that each dust tracer must follow each other in the field_table in ascending order with their size (smallest to largest)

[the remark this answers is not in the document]
jianheACM: Do we have to have the same number of emis_dust bins as the number of dust tracer bins? If so, maybe we can just use ndust for the loop.

drivers/ccpp/catchem_dust_wrapper.F90
```diff
@@ -210,10 +210,10 @@ subroutine catchem_dust_wrapper_run(im, kte, kme, ktau, dt, garea, land,   &
           do i=its,ite
             if(xland(i,j).lt.1.5)then ! JianHe: why not also > 0.5?
               delp = p8w(i,kts,j)-p8w(i,kts+1,j)
-              ! -- GOCART afwa dust scheme
+              ! -- FENGSHA dust scheme
               call gocart_dust_fengsha_driver(dt,&
                 chem(i,kts,j,:),rho_phy(i,kts,j), &
-                dz8w(i,kts,j),smois(i,:,j), &
+                dz8w(i,kts,j),smois(i,1,j), &
```
jordanschnell, line 216: Should we also allow for flexible #'s of soil levels to be passed in? (e.g., RUC with very shallow first layer)

src/gocart_dust_default_mod.F90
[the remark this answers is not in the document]
bbakernoaa: We are just using dust bins for emission and tracers here but the pointers are different.  Only FENGSHA is bin agnostic by default.  The other schemes we would need to sum the total mass to get the total emission due to the threshold velocity parameterization in AFWA and GINOUX

src/gocart_dust_default_mod.F90
```diff
@@ -2,209 +2,225 @@
...
+               SQRT(1.0+0.006/den(n)/g/(diam(n))**2.5)/ &
+               SQRT(1.928*(1331.0*(diam(n))**1.56+0.38)**0.092-1.0) 
+
+          ! Case of surface dry enough to erode
+          IF (gwet < gthresh) THEN
+             u_ts = MAX(0.0D+0,u_ts0*(1.2D+0+2.0D-1*LOG10(MAX(1.0D-3, gwet))))
+          ELSE
+             ! Case of wet surface, no erosion
+             u_ts = 100.0_kind_chem
+          END IF
```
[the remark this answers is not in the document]
bbakernoaa: Moved the parts out of the loop that are redundant

src/dust_afwa_mod.F90
```diff
@@ -25,361 +23,384 @@ subroutine gocart_dust_afwa_driver(ktau,dt,u_phy,              &
...
+    REAL(kind=kind_chem), INTENT(IN) :: v10
+    REAL(kind=kind_chem), INTENT(IN) :: delp
+    REAL(kind=kind_chem), INTENT(IN) :: area
+    REAL(kind=kind_chem), INTENT(IN) :: ust
+    REAL(kind=kind_chem), INTENT(IN) :: znt
+    REAL(kind=kind_chem), INTENT(IN) :: clay
+    REAL(kind=kind_chem), INTENT(IN) :: sand
+
+    REAL(kind=kind_chem), DIMENSION(num_chem), INTENT(INOUT) ::  chem_arr
+    REAL(kind=kind_chem), DIMENSION(ndust), INTENT(INOUT) :: emis_dust
```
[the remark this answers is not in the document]
zmoon: But do we need both then?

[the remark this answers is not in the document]
bbakernoaa: I don't think so...  need to check upstream which is set first but it could be both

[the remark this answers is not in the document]
jianheACM: The question is should the num_emis_dust  always be the number of dust tracer bins (same for seasalt). If so, we can just use ndust. If not, we need keep both.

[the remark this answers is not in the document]
bbakernoaa: I think for GOCART right now it is true.  When we start bringing in a different aerosol description we can revisit it imo.

src/dust_afwa_mod.F90
bbakernoaa: Lets handle this in a future PR

drivers/ccpp/catchem_dust_wrapper.F90
bbakernoaa: I'll handle this in a future PR

parameters/dust_data_mod.F90
```diff
@@ -39,38 +39,7 @@ module dust_data_mod
   ! -- FENGSHA
   real(kind_chem), parameter :: fengsha_alpha = 0.3
   real(kind_chem), parameter :: fengsha_gamma = 1.3
-  ! -- FENGSHA threshold velocities based on Dale A. Gillette's data
-  integer, parameter :: fengsha_maxstypes = 13
```
bbakernoaa, line 43: these aren't used anymore.  from a very old version of fengsha

parameters/dust_data_mod.F90
```diff
@@ -79,7 +48,7 @@ module dust_data_mod
   real(kind_chem) :: dust_gamma = 0.
 
 
-  ! -- sea salt parameters
+  ! -- Saltation Parameters
```
bbakernoaa, line 51: correct from sea salt to saltation

bbakernoaa:
I have a sample run on HERA here: `/scratch2/NCEPDEV/stmp3/Barry.Baker/COMROT/test2/gfs.20160715`

drivers/ccpp/catchem_dust_wrapper.F90
```diff
@@ -189,18 +190,22 @@ subroutine catchem_dust_wrapper_run(im, kte, kme, ktau, dt, garea, land,   &
...
+                v_phy(i,kts,j),rho_phy(i,kts,j), &
                 dz8w(i,kts,j),smois(i,:,j),u10(i,j), &
                 v10(i,j),delp,erod(i,j,:),isltyp(i,j),dxy(i,j),    &
                 emis_dust(i,1,j,:),&
                 ust(i,j),znt(i,j),clayf(i,j),sandf(i,j),nsoil)
               !store_arrays = .true.
-            endif
-          enddo
-        enddo
+              do n=0,ndust-1
```
jordanschnell, line 203: duplicated for each opt, maybe collapse to one at end of case select?

zmoon: The other two don't have the RHS `dxy`, I notice.

parameters/catchem_constants.mod
zmoon, line 1: There are a few module files here. Suggest removing them and adding `*.mod` to the `.gitignore`.

drivers/ccpp/catchem_dust_wrapper.F90
bbakernoaa: There is a slight difference.  I suggest we move forward to get this in and have a subsequent update this going forward.

bbakernoaa:
@rschwant @jordanschnell @jianheACM @zmoon Please take another look at this we need to move forward

jordanschnell:
> @jianheACM @jordanschnell @zmoon Please take another look at this. Fixed it up and is producing the same results.
> 
> @jordanschnell I updated the all the dust schemes to pass the tendencies out and then update in the catchem_dust_driver. Please take a look.

This part looks good to me. As I mentioned and @zmoon noted, would be good to have a single line to add the emissions but I recognize there are slight differences so all good for now.

drivers/ccpp/catchem_dust_wrapper.F90
jianheACM, line 1: @bbakernoaa Can you add a note under "!! Revision History:" for the dust updates?

bbakernoaa: do we really need those? they are all tracked by the git commits

jianheACM: I'm not saying we need to add notes for each commit, just a brief summary (a few words with your contact information, e.g., dust module updated by Barry Baker (email)).  I just feel like we have so many commits and will have many more in the future. Probably not everyone would be interested in checking each commit. In that case, users who have specific dust-related questions can directly go to you. That being said, I'm also ok if you prefer not including this.

PR #11, opened 2023-09-01
Update to Seasalt: Code Cleanup and bin invariance 

bbakernoaa:
This is a simple code cleanup to the sea salt schemes

Includes updates to sea salt to make he code a little more readable.

Also include updates to make the number of bins arbitrary within the schemes themselves.


src/gocart_seas_default_mod.F90
```diff
@@ -17,81 +17,77 @@ module gocart_seas_default_mod
...
+    ! calcualte airmass 
+    ! -----------------
     airmas=area * delp / g
-    dxy=area
     ipr=0
 
-    airmas1 = airmas
-    tc1(:) = tc
-    bems1(:) = bems
-    call source_ss( number_ss_bins, dt, tc1,ilwi, dxy, w10m, airmas1, bems1,ipr)
```
jianheACM, line 78: It should be 
call source_ss( number_ss_bins, dt, tc1,ilwi, dxy, **ws10m**, airmas1, bems1,ipr) 
This is my fault when restructring the seasalt code.

src/gocart_seas_default_mod.F90
```diff
@@ -17,81 +17,77 @@ module gocart_seas_default_mod
...
+    ! ---------------
+    INTEGER, INTENT(IN) :: ktau
+    REAL(kind=kind_chem), INTENT(IN) :: dt
+    REAL(kind=kind_chem), INTENT(IN) :: u_phy
+    REAL(kind=kind_chem), INTENT(IN) :: v_phy
+    REAL(kind=kind_chem), INTENT(IN) :: dz8w
+    REAL(kind=kind_chem), INTENT(IN) :: u10
+    REAL(kind=kind_chem), INTENT(IN) :: v10
+    REAL(kind=kind_chem), INTENT(IN) :: delp
+    REAL(kind=kind_chem), INTENT(IN) :: tsk ! this isn't used anywhere
```
zmoon, line 33:
```suggestion
    REAL(kind=kind_chem), INTENT(IN) :: tsk ! FIXME: this isn't used anywhere
```
helpful to use such comment tags that can later be easily found

src/gocart_seas_default_mod.F90
```diff
@@ -17,81 +17,77 @@ module gocart_seas_default_mod
...
+
+    ! Calcualte 10m from first model layer if first layer is low
     if (dz8w < 12.) then
       ws10m = sqrt(u_phy*u_phy+v_phy*v_phy)
     else
       ws10m = sqrt(u10*u10+v10*v10)
     end if
 
-    ilwi=0
+    ilwi=0 ! this is always zero is it needed?
```
zmoon, line 74: Indeed seems like it would be better to eliminate it and eliminate that corresponding else branch.

src/gocart_seas_ngac_mod.F90
```diff
@@ -121,18 +126,12 @@ subroutine gocart_seas_ngac(ktau,dt,u_phy,              &
...
-      chem_arr(p_seas_4) = chem_arr(p_seas_4) + tc(4)*converi
-      chem_arr(p_seas_5) = chem_arr(p_seas_5) + tc(5)*converi
-
-      ! for output diagnostics kg/m2/s
-      emis_seas(p_eseas1) = bems(1)
-      emis_seas(p_eseas2) = bems(2)
-      emis_seas(p_eseas3) = bems(3)
-      emis_seas(p_eseas4) = bems(4)
-      emis_seas(p_eseas5) = bems(5)
+      do n = 0, number_ss_bins -1
```
zmoon, line 129: Seems like this assumes that `number_ss_bins` and `num_emis_seas` are the same. So do we need both?

bbakernoaa: I don't think so

PR #114, opened 2025-01-27
Incorporate MICM into build and start chem process structure

zmoon:
cc: @mattldawson


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.5)
 
-project(CATChem VERSION 0.1.0 LANGUAGES Fortran)
+project(CATChem VERSION 0.1.0 LANGUAGES Fortran C CXX)
```
zmoon, line 3: Just nothing that, without adding these, was getting this error:
```
CMake Error in src/process/chem/CMakeLists.txt:
  No known features for CXX compiler

  ""

  version .
```

src/external/CMakeLists.txt
```diff
@@ -13,9 +17,32 @@ endif()
...
+FetchContent_Declare(
+  musica_fortran
+  GIT_REPOSITORY https://github.com/NCAR/musica.git
+  GIT_TAG v0.9.0
+)
+
+set(MICM_ENABLE_TESTS OFF)
+set(MICM_ENABLE_EXAMPLES OFF)
+
+set(MUSICA_BUILD_C_CXX_INTERFACE ON)
```
zmoon, line 41: @mattldawson @K20shores I saw that [this example](https://github.com/NCAR/musica/blob/e0c3d219fc8bbfd388cd3c95a3646d2e3aa1e288/fortran/test/fetch_content_integration/CMakeLists.txt#L22) has it off, but that gave me this error:

```
-- CMake build configuration for musica-fortran (Debug) 0.9.0
--   MICM: ON
--   TUV-x: OFF
--   Tests: OFF
--   Install: OFF
--   C/C++ Interface: OFF
CMake Error at build/_deps/musica_fortran-src/fortran/CMakeLists.txt:23 (find_package):
  By not providing "Findmusica.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "musica", but
  CMake did not find one.

  Could not find a package configuration file provided by "musica" with any
  of the following names:

    musicaConfig.cmake
    musica-config.cmake

  Add the installation prefix of "musica" to CMAKE_PREFIX_PATH or set
  "musica_DIR" to a directory containing one of the above files.  If "musica"
  provides a separate development package or SDK, be sure it has been
  installed.
```

CMakeLists.txt
zmoon: Also note, it seems like really it's only necessary to add CXX to get a successful build

src/external/CMakeLists.txt
K20shores: Thanks for tagging me. I'll try to help troubleshoot this.

tests/test_chem.f90
```diff
@@ -0,0 +1,14 @@
+program test_chem
+   use CATChem, only: cc_get_micm_version
+   use testing_mod, only: assert
+   implicit none
+
+   character(len=:), allocatable :: micm_version
+   character(len=*), parameter :: expected_micm_version = "3.7.0"
+
+   micm_version = adjustl(trim(cc_get_micm_version()))
```
zmoon: The `adjustl` isn't really needed, but to work properly (as intended) it should be in this order:
```suggestion
   micm_version = trim(adjustl(cc_get_micm_version()))
```

CMakeLists.txt
zmoon: @K20shores do you know if there's a way to get MUSICA to build without changing this, by specifying something elsewhere instead? Just curious, as our project code is all Fortran really.

src/external/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+#
+# GOCART2G Process Library
```
zmoon:
```suggestion
# GOCART-2G Process Library
```

bbakernoaa (approved):

bbakernoaa:
@mbruckner-work Please take a look

rschwant:
Very cool! Let's talk about this in our UFS-Chem meetings on Monday. I have a couple of questions, but this looks great!

CMakeLists.txt
K20shores: @zmoon I think this is required. If not, I don't think the linker will be properly configured. I could be wrong though

src/process/chem/CCPr_Chem_Mod.F90
jianheACM, line 1: Is it currently a placeholder here or it has already been fully connected to micm?

zmoon: It's more of a placeholder for other MICM stuff. We'll eventually have to add similar to things to what Matt has showed for CAM-SIMA.

jianheACM: Sounds good!

jianheACM (approved):

CMakeLists.txt
K20shores: @zmoon I was able to build your fork

zmoon: Thanks @K20shores , that's good to hear.

src/external/CMakeLists.txt
mattldawson: I think you need to have this turned on unless you have already installed the MUSICA library (the one in C). All the wrapper libraries depend on the MUSICA library.

The reason we made this an option was to allow you to compile the MUSICA library and the MUSICA-Fortran libraries with different compilers (e.g., see [here](https://github.com/NCAR/musica/blob/e0c3d219fc8bbfd388cd3c95a3646d2e3aa1e288/docker/Dockerfile.fortran-intel#L33-L72)). This was needed to support using the LLVM libraries for JIT compiling in MICM (which only works with gcc and clang), but it's looking like we may not actually need to do JIT compiling to achieve our performance goals, so this option may go away in the future.

zmoon: Interesting, thanks!

src/process/chem/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(_lib CATChem_process_chem)
+
+set(_srcs CCPr_Chem_Mod.F90)
+
+add_library(${_lib} ${_srcs})
```
K20shores, line 5: The only cmake related comment I can think of is that it's recommended to use [`target_sources`](https://cmake.org/cmake/help/latest/command/target_sources.html) in newer versions of cmake. I think if you use `target_sources` and generate a project for something like Xcode or Visual Studio, this helps Cmake create better integration with that IDE. This certainly isn't a requirement though

CMakeLists.txt
K20shores: Ah, if you want to only enable these languages only when musica is built, somewhere you can do something like this

```
if (ENABLE_MUSICA)
    enable_language(C)
    enable_language(CXX)
endif()
```

K20shores (approved):

src/process/chem/CMakeLists.txt
zmoon: Thanks @K20shores , I can update the whole project to use that in a separate PR.

mattldawson (approved):
looks great!

PR #117, opened 2025-03-19
Add GOCART SUV process

lwcugb:
This is an interface to  GOCART Sulfur Volcanic Emissions of SO2. The PR is an update to the previous one #113


zmoon:
@lwcugb could you run `pre-commit run --all-files` and commit the result? Then the linter should pass.

Also, the assertion in the test failed with `gfortran-12` (and signaled underflow: `IEEE_UNDERFLOW_FLAG` `IEEE_DENORMAL`), but interestingly not with `gfortran-10`. So we should examine what's in that array.

lwcugb:
@zmoon Thanks for checking it. I made a new commit to fix the lint error as you suggested. The test works fine on my local machine with gfortran-14 too. Not sure what is special for gfortran-12. Any idea how I should fix it?

zmoon:
> The test works fine on my local machine with gfortran-14 too. Not sure what is special for gfortran-12.

The `IEEE_UNDERFLOW_FLAG`/`IEEE_DENORMAL` seem to suggest that the calculations are producing numbers too small to normally represent with the selected precision (*subnormal* numbers), ending up in the `SUVolcanicState%TotalEmission` array. It seems that the `gfortran-12` case is flushing those to zero, such that the `sum(...) > 0` is not true.

I would print out the array values you get with your compiler, see if they are the magnitude you expect. And I will test with `gfortran-12`.

Edit: in the latest run, `gfortran-10` and `gfortran-11` failed, while `gfortran-10` was working before. Perhaps hardware differences in the runners used for the different runs?

lwcugb:
The non-zero values in the array are not too small on a magnitude of `e-3`. I have way smaller values for BVOC emissions to a magnitude of `e-11`, which works fine. I am guessing there might be too many zeros in the array. I only have emissions at 4 levels out of 28. So I changed the assertion to `ANY(SUVolcanicState%TotalEmission > 0.0_fp)` .

zmoon:
Looks like the issue was some missing initializations, so parts of the array were getting random garbage. Here's how you could fix that:

<details><summary>diff</summary>

```diff
--- a/src/process/SUVolcanicEmissions/CCPr_SUVolcanicEmissions_Mod.F90
+++ b/src/process/SUVolcanicEmissions/CCPr_SUVolcanicEmissions_Mod.F90
@@ -134,11 +134,13 @@ CONTAINS
          ALLOCATE( SUVolcanicState%SUVolcanicSpeciesIndex(SUVolcanicState%nSUVolcanicSpecies), STAT=RC )
          CALL CC_CheckVar('SUVolcanicState%SUVolcanicSpeciesIndex', 0, RC)
          IF (RC /= CC_SUCCESS) RETURN
+         SUVolcanicState%SUVolcanicSpeciesIndex = -1

-         ! Allocate emission speceis names
+         ! Allocate emission species names
          ALLOCATE( SUVolcanicState%SUVolcanicSpeciesName(SUVolcanicState%nSUVolcanicSpecies), STAT=RC )
          CALL CC_CheckVar('SUVolcanicState%SUVolcanicSpeciesName', 0, RC)
          IF (RC /= CC_SUCCESS) RETURN
+         SUVolcanicState%SUVolcanicSpeciesName = ''

          ! Allocate CatChem species index
          ALLOCATE( SUVolcanicState%SpcIDs(SUVolcanicState%nSUVolcanicSpecies), STAT=RC )
@@ -150,11 +152,13 @@ CONTAINS
             SIZE(EmisState%Cats(SUVolcanicState%CatIndex)%Species(1)%Flux)), STAT=RC )
          CALL CC_CheckVar('SUVolcanicState%EmissionPerSpecies', 0, RC)
          IF (RC /= CC_SUCCESS) RETURN
+         SUVolcanicState%EmissionPerSpecies = ZERO

          ! Allocate total emissions
          ALLOCATE( SUVolcanicState%TotalEmission(SIZE(EmisState%Cats(SUVolcanicState%CatIndex)%Species(1)%Flux)), STAT=RC )
          CALL CC_CheckVar('SUVolcanicState%TotalEmission', 0, RC)
          IF (RC /= CC_SUCCESS) RETURN
+         SUVolcanicState%TotalEmission = ZERO

          ! Set the file directory
          SUVolcanicState%FileDir = TRIM(Config%suvolcanic_filedir)
@@ -261,6 +265,7 @@ CONTAINS
                !set area and SO2
                allocate(area(1,1), SO2(1,1, MetState%NLEVS))
                area(1,1) = MetState%AREA_M2
+               SO2 = ZERO

                ! loop through all species. Right now, GOCART only has SO2
                do i = 1, SUVolcanicState%nSUVolcanicSpecies
diff --git a/tests/test_suvolcanic.F90 b/tests/test_suvolcanic.F90
index 7df7d39..79fd8d5 100644
--- a/tests/test_suvolcanic.F90
+++ b/tests/test_suvolcanic.F90
@@ -118,7 +118,7 @@ program test_suvolcanic
       stop 1
    end if

-   call assert( ANY(SUVolcanicState%TotalEmission > 0.0_fp), "Test Sulfur Volcanic Emissions")
+   call assert(sum(SUVolcanicState%TotalEmission) > 1.0e-2_fp, "Test Sulfur Volcanic Emissions")
    call print_info(Config, SUVolcanicState, MetState, title)
```
</details>

You could also assert that all the levels that aren't the four that have values are zero (which is what we want here).

> magnitude of `e-3`. I have way smaller values for BVOC emissions to a magnitude of `e-11`

Is this a units issue or difference in units that needs to be addressed?

lwcugb:
Thanks @zmoon for the fix. It depends on the dummy MET inputs and the volcano heights to determine which level has non-zero values. So it can change. I keep using `sum(SUVolcanicState%TotalEmission) > 0.0_fp` because I am not sure if `1.0e-2_fp` is the minimum total emissions, which agian depends on inputs. 

I guess the magnitude between emission categories can be quite different. It might be due to the fact that I am using all the volcanoes to this one grid in the test, which should not be true in real run.

zmoon:
> It depends on the dummy MET inputs and the volcano heights to determine which level has non-zero values. So it can change. I keep using `sum(SUVolcanicState%TotalEmission) > 0.0_fp` because I am not sure if `1.0e-2_fp` is the minimum total emissions, which agian depends on inputs.

In this test case, the inputs are fixed, so it's okay to be more specific, e.g. including magnitude. If you make a code change to SUV or your input settings that causes the emissions magnitude to change, you just update the assertion. Similarly, currently, every time you run this test case, the levels that have data should be the same (seems to be 4--7).

lwcugb:
Ok. So I added assertions as you suggested and made a comment to be very specific about them.

CMakeLists.txt
```diff
@@ -54,5 +54,6 @@ elseif(
   )
 endif()
 
+add_subdirectory(util)
```
zmoon:
```suggestion
```
Do you think we need any of that stuff? If not we could remove this line and the util subdir.

src/process/SUVolcanicEmissions/CCPr_SUVolcanicEmissions_Mod.F90
```diff
@@ -0,0 +1,382 @@
...
+      INTEGER                         :: SchemeOpt             ! Scheme Option (if there is only one SchemeOpt always = 1)
+      integer                         :: nSUVolcanicSpecies           !< Number of SUVolcanic species
+      integer, pointer                :: SUVolcanicSpeciesIndex(:)    !< Index of SUVolcanic species
+      character(len=31), pointer      :: SUVolcanicSpeciesName(:)     !< name of SUVolcanic species
+      integer, pointer                :: SpcIDs(:)               !< CATChem species IDs
+      integer                         :: CatIndex                !< Index of emission category in EmisState
+
+      ! Process Specific Parameters
+      real(fp), pointer               :: TotalEmission(:)          !< Total emission of all species at each level [kg/m^2/s]
+      real(fp), pointer               :: EmissionPerSpecies(:,:)   !< Emission per species at each level          [kg/m^2/s]
```
zmoon: Can you change these `pointer` ones to be normal `allocatable` arrays instead?

lwcugb: Done in the new commit!

CMakeLists.txt
lwcugb: Yes. There is a function to read the volcanic emission files and convert MET data to be consistent with the GOCART formant. If we continue adding more processes from GACART, they can be pretty handy.

zmoon: How big are those files? Can we write a test of that code?

lwcugb: Each daily emission file is small. You can find an example used in this test under `/tests/Emissions/so2_volcanic_emissions_Carns.20220101.rc`. It only contains one subroutine in the `reademissions.F90` module. I used it in my test and it works fine. I am not sure if an additional test is necessary.

zmoon: I meant a test that is part of our test suite here that directly uses the code in the new `src/util/` modules. But looking closer it looks like SUV is using those routines. So I would suggest we move those routines `ccpr_scheme_GOCART_SUVolcanicEmissions_mod`. They shouldn't be outside of our `src/` tree.

lwcugb: Agreed! Lacey added other simple functions in the DMS pull request #115. They are to convert MET to gocart [format](https://github.com/ufs-community/CATChem/pull/115/files#diff-6fc14b343ee23033f56765a8e413917480252e2e3e22d5300472a2c43241ec85R60) . I think they can be applied to other Gocart processes too. Do you think they should be moved too?

src/process/CMakeLists.txt
```diff
@@ -4,3 +4,4 @@ add_subdirectory(dust)
 add_subdirectory(seasalt)
 add_subdirectory(plumerise)
 add_subdirectory(drydep)
+add_subdirectory(SUVolcanicEmissions)
```
zmoon:
```suggestion
add_subdirectory(volcanic)
```
More consistent with the others? If you agree, should change the module name too.

CMakeLists.txt
zmoon: If there are routines that work for both processes, we may want to think about another place to put them, but it would still be in the `src/` tree (probably sitting within `src/process/`). I think for this PR, though, let's just move them to our GOCART SUV mod for now.

src/process/CMakeLists.txt
lwcugb: Done!

CMakeLists.txt
lwcugb: Done!

src/process/volcanic/ccpr_scheme_Volcanic_GOCART_mod.F90
zmoon, line 1: You should be able to replace the rest of the `pointer` cases here with `allocatable` arrays.

util/get-vars.py
zmoon, line 1: This file was not related to this PR, please put it back (I can do it if you have issue).

util/vars.yml
zmoon, line 1: This file was not related to this PR, please put it back (I can do it if you have issue).

util/metutils_mod.F90
zmoon, line 1: This one might be okay to remove, but we should check with @bbakernoaa

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+      CHARACTER(LEN=18) :: ymd_str          ! Model date in string
+
+      INTEGER, dimension(:), allocatable   :: vStart      ! Emissions Start time [sec]
+      INTEGER, dimension(:), allocatable   :: vEnd        ! Emissions end time [sec]
+      INTEGER                              :: nVolc       ! number of volcanic sources
+      INTEGER, dimension(:), allocatable   :: iPoint, jPoint ! sub-domain - we only run this at the place/time of eruption??
+      INTEGER, parameter                   :: nSO2 =1     ! index of SO2 relative to other sulfate tracers
+
+      REAL, dimension(:), allocatable      :: vSO2   ! volcanic emissions  [kg]
+      REAL, dimension(:,:,:),pointer  :: SO2       ! SO2 [kg kg-1]
```
zmoon, line 217: This one too

util/get-vars.py
lwcugb: Sorry, I thought all the files in util are not needed. Could you help me with that because I am not aware of a quick way to do it.

src/process/volcanic/CCPr_Volcanic_Mod.F90
lwcugb: These pointers are kept so that they can be used in GOCART. That is the format defined in GOCART. I am not sure if it will give an error if they are changed to allocatables.

zmoon: You should be able to pass allocatable to GOCART where it says it wants pointer. In this situation the allocatable array serves the same function but it's a little safer. I tried it out at some point. But let me know if you have issue.

util/get-vars.py
zmoon: Done!

zmoon:
@lwcugb I also merged in the update to GOCART submodule. Not sure if anything affects your work here, but here's the changelog: https://github.com/GEOS-ESM/GOCART/blob/v2.4.0/CHANGELOG.md

lwcugb:
> @lwcugb I also merged in the update to GOCART submodule. Not sure if anything affects your work here, but here's the changelog: https://github.com/GEOS-ESM/GOCART/blob/v2.4.0/CHANGELOG.md

Thanks for the heads up. I do have to make some minor changes and it works again.

src/process/volcanic/CCPr_Volcanic_Mod.F90
zmoon: 🤷

src/process/volcanic/ccpr_scheme_Volcanic_GOCART_mod.F90
bbakernoaa, line 1: typically the file names have all been lowercase.  Could you change the filename to follow the convention.

bbakernoaa (changes requested):
Just that scheme filename to be changed

bbakernoaa (approved):

bbakernoaa:
@lwcugb this should be closed now right?

PR #120, opened 2025-03-27
Add GOCART DMS process

lwcugb:
Adding GOCART DMS scheme based on PR #115 

Note that this branch builds on SUV PR #117, and it should be merged first.


bbakernoaa (changes requested):
same with the file name request here.  `ccpp_scheme_GOCART_DMS_mod.F90 -> ccpr_scheme_gocart_dms_mod.F90`

src/core/chemstate_mod.F90
```diff
@@ -40,13 +40,15 @@ module ChemState_Mod
...
    !! \param SpeciesIndex: An array containing the total species index.
    !! \param AeroIndex: An array containing the aerosol species index.
    !! \param GasIndex: An array containing the gas species index.
    !! \param DustIndex: An array containing the dust species index.
+   !! \param DryDepIndex: An array containing the dry deposition species index.
    !! \param SeaSaltIndex: An array containing the sea salt species index.
-   !! \param chemSpecies: A 2-D array containing the concentration of each species.
+   !! \param DryDepIndex: An array containing the dry deposition species index.
+   !! \param SpeciesNames: A character array containing the names of the species.
    !!
```
rschwant, line 51: Looks like DryDepIndex is listed twice here. For clarity, should we remove one of them?

src/process/dms/CCPr_DMS_Mod.F90
```diff
@@ -0,0 +1,299 @@
...
+   !!
+   !! \details Contains all the information needed to run DMS Process
+   !!
+   !! This type contains the following variables:
+   !! \param Activate Activate Process (True/False)
+   !! \param SchemeOpt Scheme Option
+   !! \param nDMSSpecies Number of DMS species
+   !! \param DMSSpeciesIndex Index of DMS species
+   !! \param DMSSpeciesName Name of DMS species
+   !! \param SpcIDs CATChem species IDs
```
rschwant, line 38: What is this SpcIDs term? I see why you need the DMS number of species, index of DMS species, and name of DMS species. But what exactly are we storing as SpcIDs?

src/process/dms/CCPr_DMS_Mod.F90
```diff
@@ -0,0 +1,299 @@
...
+
+   TYPE :: DMSStateType
+      LOGICAL                         :: Activate              ! Activate Process (True/False)
+      INTEGER                         :: SchemeOpt             ! SchemeOption
+      integer                         :: nDMSSpecies           !< Number of DMS species
+      integer, allocatable            :: DMSSpeciesIndex(:)    !< Index of DMS species
+      character(len=31), allocatable  :: DMSSpeciesName(:)     !< name of DMS species
+      integer, allocatable            :: SpcIDs(:)             !< CATChem species IDs
+      integer                         :: CatIndex              !< Index of emission category in EmisState
+
```
rschwant, line 54: Why are some of those above lowercase and some capitalized? I don't think it matters in Fortran, but best to be consistent. Most of the above is capitalized, can we make these capitalized by convention too?

src/process/dms/CCPr_DMS_Mod.F90
```diff
@@ -0,0 +1,299 @@
...
+      USE CCPr_Scheme_GOCART_DMS_Mod, only : CCPR_Scheme_GOCART_DMS
+
+      IMPLICIT NONE
+
+      ! INPUT PARAMETERS
+      TYPE(MetStateType),  INTENT(IN)      :: MetState     ! MetState Instance
+
+      ! INPUT/OUTPUT PARAMETERS
+      TYPE(DMSStateType), INTENT(INOUT)    :: DMSState     ! DMSState Instance
+      TYPE(EmisStateType),  INTENT(INOUT)  :: EmisState    ! ChemState Instance
```
rschwant: Should this comment really be EmisState Instance instead?

src/process/dms/CCPr_DMS_Mod.F90
```diff
@@ -0,0 +1,299 @@
...
+
+      ! OUTPUT PARAMETERS
+      INTEGER, INTENT(OUT)                 :: RC           ! Return Code
+
+      ! LOCAL VARIABLES
+      INTEGER :: s
+      CHARACTER(LEN=255) :: ErrMsg, thisLoc
+      REAL, dimension(:,:,:),pointer  :: SU_emis   ! SU emissions, kg/m2/s
+
```
rschwant, line 193: Can we define SU here in the comment?

src/process/dms/CCPr_DMS_Mod.F90
```diff
@@ -0,0 +1,299 @@
...
+      if (DMSState%Activate) then
+         ! Run the DMS Scheme
+         !-------------------------
+         do s = 1, DMSState%nDMSSpecies !only one species for now
+
+            if (DMSState%SchemeOpt == 1) then
+
+               !NDMS = 1 is moved to scheme module
+               allocate(SU_emis(1,1,1)); SU_emis = ZERO
+
```
rschwant, line 211: What are these dimensions representing? Can we add a comment for this?

src/process/dms/ccpr_scheme_GOCART_DMS_mod.F90
```diff
@@ -0,0 +1,156 @@
...
+   !! \param cdt          model timestep [sec]
+   !! \param g0           gravity [m/s2]
+   !! \param tmpu         Temperature [K]
+   !! \param u10m         10-m u-wind component [m/sec]
+   !! \param v10m         10-m v-wind component [m/sec]
+   !! \param lwi          orography flag; Land, ocean, ice mask
+   !! \param delp         Pressure Thickness for layer [Pa]
+   !! \param dmso_conc    DMS source concentration [nmol/L]
+   !! \param SU_emis      SU emissions, kg/m2/s
+   !! \param ndms         index of DMS relative to other sulfate tracers
```
rschwant: This is the only local variable defined here. Should this include local variable definitions too? And if so, for consistency, should we add fMassDMS too and others? Or is it best to not include local variables here?

src/process/dms/ccpr_scheme_GOCART_DMS_mod.F90
```diff
@@ -0,0 +1,156 @@
...
+      INTEGER, intent(in) :: lwi               ! orography flag; Land, ocean, ice mask
+      REAL, dimension(:), intent(in) :: delp   ! Pressure Thickness for layer [Pa]
+      REAL, intent(in)  :: dmso_conc           ! DMS source concentration [nmol/L]
+      REAL, intent(inout),dimension(:,:,:),pointer  :: SU_emis   ! SU emissions, kg/m2/s
+      integer, intent(out) :: RC               ! Success or Failure
+
+      ! Local Variables
+      INTEGER, parameter :: NDMS = 1    ! index of DMS relative to other sulfate tracers
+      REAL, parameter :: fMassDMS=62.   ! DMS molecular weight [g/mol]
+      character(len=256) :: errMsg
```
rschwant: Can we confirm that the MW of DMS in GOCART only uses 62. as opposed to more precision? We should eventually have the MW stored in one place of the code and update this to refer to the mechanism files, but this can be done later.

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+   !! \param Scheme Scheme Option
+   !! \param nVolcanicSpecies # of Volcanic species
+   !! \param VolcanicSpeciesIndex Index of Volcanic species
+   !! \param VolcanicSpeciesName Name of Volcanic species
+   !! \param SpcIDs CATChem species IDs
+   !! \param CatIndex Index of emission category in EmisState
+   !! \param TotalEmission Total emission of all species at each level [kg/m^2/s]
+   !! \param EmissionPerSpecies Emission per species at each level [kg/m^2/s]
+   !! \param FileDir Input file directory for reading in emissions
+   !!
```
rschwant, line 41: Let's update this group of comments to include more spacing between the parameter name and description. Also Scheme looks to actually be SchemeOpt below.

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+      integer, allocatable            :: VolcanicSpeciesIndex(:)    !< Index of Volcanic species
+      character(len=31), allocatable  :: VolcanicSpeciesName(:)     !< name of Volcanic species
+      integer, allocatable            :: SpcIDs(:)               !< CATChem species IDs
+      integer                         :: CatIndex                !< Index of emission category in EmisState
+
+      ! Process Specific Parameters
+      real(fp), allocatable           :: TotalEmission(:)          !< Total emission of all species at each level [kg/m^2/s]
+      real(fp), allocatable           :: EmissionPerSpecies(:,:)   !< Emission per species at each level          [kg/m^2/s]
+      character(len=1055)             :: FileDir                  !< Input file directory for reading in emissions
+
```
rschwant, line 59: Again I would update this to have all capitals so that it looks more consistent. It looks messy to have both INTEGER and integer for example in the same place of the code.

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+   SUBROUTINE CCPR_Volcanic_Init( Config, VolcanicState, EmisState, RC )
+      ! USE
+
+
+      IMPLICIT NONE
+      ! INPUT PARAMETERS
+      !-----------------
+      TYPE(ConfigType)    :: Config    ! Module options
+      TYPE(EmisStateType) :: EmisState ! Chemical state
+
```
rschwant, line 86: I think the comment here should be Emis state and not Chemical state

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+      IMPLICIT NONE
+      ! INPUT PARAMETERS
+      TYPE(MetStateType),  INTENT(IN) :: MetState       !< MetState Instance
+
+      ! INPUT/OUTPUT PARAMETERS
+      !TYPE(DiagStateType), INTENT(INOUT)      :: DiagState       !< DiagState Instance
+      TYPE(VolcanicStateType), INTENT(INOUT)  :: VolcanicState  !< VolcanicState Instance
+      !TYPE(ChemStateType), INTENT(INOUT)     :: ChemState       !< ChemState Instance
+      TYPE(EmisStateType), INTENT(INOUT)     :: EmisState       !< ChemState Instance
+
```
rschwant, line 196: Would we ever need the DiagState or ChemState here? If not, we should delete for clarity.

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+      ! LOCAL VARIABLES
+      CHARACTER(LEN=255) :: ErrMsg, thisLoc
+      CHARACTER(LEN=1055) :: fname
+      type(VolcanicEmisData), allocatable :: VolcanicEmis(:)
+      CHARACTER(len=7), parameter :: label='volcano'
+      INTEGER :: i                          ! loop index
+      INTEGER :: hms                        ! Model time [secs] TODO: format is right?
+      INTEGER :: ymd                        ! Model date [YYYYMMDD] TODO: format is right?
+      CHARACTER(LEN=18) :: ymd_str          ! Model date in string
```
rschwant, line 208: What do these TODO notes mean? Do we need to confirm this still?

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+      INTEGER, dimension(:), allocatable   :: iPoint, jPoint ! sub-domain - we only run this at the place/time of eruption??
+      INTEGER, parameter                   :: nSO2 =1     ! index of SO2 relative to other sulfate tracers
+
+      REAL, dimension(:), allocatable      :: vSO2   ! volcanic emissions  [kg]
+      REAL, dimension(:,:,:),pointer  :: SO2       ! SO2 [kg kg-1]
+      !REAL, dimension(:,:,:),pointer  :: SU_emis   ! SU emissions, kg/m2/s
+      REAL, dimension(:), allocatable        :: vCloud    ! top elevation of emissions [m]
+      REAL, dimension(:), allocatable        :: vElev     ! bottom elevation of emissions [m]
+      REAL, dimension(:), allocatable        :: vLat     ! latitude specified in file [degree]
+      REAL, dimension(:), allocatable        :: VLon     ! longitude specified in file [degree]
```
rschwant: It looks like this should really be vLon to be consistent with that below and the other tracers.

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+
+      INTEGER, dimension(:), allocatable   :: vStart      ! Emissions Start time [sec]
+      INTEGER, dimension(:), allocatable   :: vEnd        ! Emissions end time [sec]
+      INTEGER                              :: nVolc       ! number of volcanic sources
+      INTEGER, dimension(:), allocatable   :: iPoint, jPoint ! sub-domain - we only run this at the place/time of eruption??
+      INTEGER, parameter                   :: nSO2 =1     ! index of SO2 relative to other sulfate tracers
+
+      REAL, dimension(:), allocatable      :: vSO2   ! volcanic emissions  [kg]
+      REAL, dimension(:,:,:),pointer  :: SO2       ! SO2 [kg kg-1]
+      !REAL, dimension(:,:,:),pointer  :: SU_emis   ! SU emissions, kg/m2/s
```
rschwant: Would you ever need this line? If not, let's delete it.

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+               vElev = VolcanicEmis(:)%Vbase
+               vLon = VolcanicEmis(:)%Vlon
+               vLat = VolcanicEmis(:)%Vlat
+               if (allocated(VolcanicEmis)) deallocate(VolcanicEmis)
+
+               !TODO: iPoint and jPoint needs to be determined; currently gives all one. In real run,
+               !we could pre-select sources for the current grid cell so giving all ones can work fine.
+               allocate(iPoint(nVolc), jPoint(nVolc))
+               iPoint = 1; jPoint = 1
+               !TODO: Not sure how to get Vstart and VEnd format and values.
```
rschwant, line 258: Can you update this comment and mention where the 235959 number comes from.

src/process/volcanic/CCPr_Volcanic_Mod.F90
```diff
@@ -0,0 +1,381 @@
...
+                  VolcanicState%TotalEmission(:) = VolcanicState%TotalEmission(:) + VolcanicState%EmissionPerSpecies(i,:)
+
+               end do ! do i = 1, VolcanicState%nVolcanicSpecies
+
+            endif  ! if (VolcanicState%nVolcanicSpecies  > 0)
+
+         endif  ! if (VolcanicState%SchemeOpt == 1)
+
+         write(*,*) 'TODO: Need to figure out how to add back to the chemical species state '
+
```
rschwant, line 312: Can we write down stuff like this that still needs to be done in all these routines? Maybe make a google doc for this.

src/process/volcanic/ccpr_scheme_Volcanic_GOCART_mod.F90
```diff
@@ -0,0 +1,300 @@
...
+   !!
+   !! \param vlat Volcano latitude
+   !! \param vlon Volcano longitude
+   !! \param VEmis Volcanic emissions [kg S/s]
+   !! \param vbase Bottom elevation of emissions [m]
+   !! \param vtop Top elevation of emissions [m]
+   !! \param nPts Number of volcanic sources in the current file
+   !! \param emissfile Emissions file name
+   !! \param label Label for emissions
+   !!
```
rschwant: Can we indent the description on these comments like that below so is cleaner and more consistent.

src/process/volcanic/ccpr_scheme_Volcanic_GOCART_mod.F90
```diff
@@ -0,0 +1,300 @@
...
+   !! \param jPoint      Grid cell index j of each volcanic source
+   !! \param hms         Current model time [sec]
+   !! \param g0          Gravity [m/s^2]
+   !! \param zbox        Geopotential Height difference [m] for layer
+   !! \param delp        Pressure Thickness for layer [Pa]
+   !! \param area        Area of grid cell [m^2]
+   !! \param vSO2        Volcanic emissions  [kg S/s]
+   !! \param nSO2        Index of SO2 relative to other sulfate tracers
+   !! \param SO2         SO2 emissions [kg kg-1]
+   !! \param SU_emis     SU emissions, kg/m2/s
```
rschwant: Why is SU_emis the only local variable defined here? Should we define all variables or leave out all local variables?

src/process/volcanic/ccpr_scheme_Volcanic_GOCART_mod.F90
```diff
@@ -0,0 +1,300 @@
...
+      REAL, intent(inout),dimension(:)        :: vElev     ! bottom elevation of emissions [m]
+      REAL, intent(inout),dimension(:)        :: vLat     ! latitude specified in file [degree]
+      REAL, intent(inout),dimension(:)        :: VLon     ! longitude specified in file [degree]
+      INTEGER, intent(inout)                  :: rc          ! error code
+
+      !local variables
+      REAL, dimension(:,:,:),pointer  :: SU_emis   ! SU emissions [kg/m2/s]
+      REAL, dimension(:,:),pointer    :: SO2EMVol  ! volcanic emissions [kg m-2 s-1]
+      REAL, parameter :: fMassSulfur = 32.  !  gram molecular weights of species
+      REAL, parameter :: fMassSO2 = 64.     !  gram molecular weights of species
```
rschwant: Again eventually need a place to define these all within the mechanism rather than here. Does GOCART only use low precision MWs fro these or should we add higher precision here?

tests/Configs/Default/CATChem_emission.yml
```diff
@@ -22,3 +22,13 @@ dust:
   DU10:
     long_name: "Dust Species Coarse Mode"
     map: ["dust1", "dust2"]
```
rschwant, line 24: What do you think about adding the MW in these yaml files instead of hardcoded in the code?

src/process/dms/CCPr_DMS_Mod.F90
lwcugb: It has not been assigned any values yet, but it is supposed to be a mapping from the DMS index to the CATChem species index. It should be useful when converting emissions to concentrations in the future.

src/process/dms/CCPr_DMS_Mod.F90
lwcugb: I guess we are following the camel naming convention here, which is also the case for other processes.

src/process/dms/ccpr_scheme_GOCART_DMS_mod.F90
lwcugb: We should only include the input and output variables to the subroutine here. `ndms` should not be here and has been deleted now.

src/process/dms/ccpr_scheme_GOCART_DMS_mod.F90
lwcugb: It is 62 indeed. The MW of DMS in GOCART is defined [here](https://github.com/GEOS-ESM/GOCART/blob/c393e57d4472c06ec61b770f8f46a152fd3280ff/ESMF/GOCART2G_GridComp/SU2G_GridComp/SU2G_GridCompMod.F90#L33). We do have MW in the species yaml file, but I am not sure if it should also be added to the emission yaml file.

src/process/volcanic/CCPr_Volcanic_Mod.F90
lwcugb: Yes. I am not sure if the format of date and time is correct for real model runs. It only works in this test.

src/process/volcanic/CCPr_Volcanic_Mod.F90
lwcugb: That is a good idea. Right now I am using TODO to record these things.

src/process/volcanic/ccpr_scheme_Volcanic_GOCART_mod.F90
lwcugb: `SU_emis` should not be here. It is deleted now.

src/process/volcanic/ccpr_scheme_Volcanic_GOCART_mod.F90
lwcugb: Yes. GOCART uses lower precision. We should be able to use higher precision if we like.

tests/Configs/Default/CATChem_emission.yml
lwcugb: It should work if MW is needed for most of the emission species. Right now, we only have DMS and sulfur. The species yaml file already has MW included, although it can be different from the emission species.

bbakernoaa:
@lwcugb are all the comments resolved?

lwcugb:
> @lwcugb are all the comments resolved?

Yes. But as suggested, we can consider adding emission species MW in the yaml too as suggested.

rschwant:
Let's update the hardcoded MWs to be based on the chemstate instead.

src/process/dms/ccpr_scheme_GOCART_DMS_mod.F90
bbakernoaa: Yes it shouldn't be in the emission yaml but rather take it from the species yaml

lwcugb: Done!

lwcugb:
> Let's update the hardcoded MWs to be based on the chemstate instead.

Done!

src/process/volcanic/ccpr_scheme_volcanic_gocart_mod.F90
```diff
@@ -0,0 +1,299 @@
...
+      !local variables
+      REAL, dimension(:,:,:),pointer  :: SU_emis   ! SU emissions [kg/m2/s]
+      REAL, dimension(:,:),pointer    :: SO2EMVol  ! volcanic emissions [kg m-2 s-1]
+      REAL, parameter :: fMassSulfur = 32.  !  gram molecular weights of species
+      REAL, parameter :: fMassSO2 = 64.     !  gram molecular weights of species
+      real, pointer :: GOCART_ZBOX(:,:,:)
+      real, pointer :: GOCART_DELP(:,:,:)
+      character(len=256) :: errMsg
+      character(len=256) :: thisLoc
+
```
rschwant, line 132: Do we need to update these too? I guess this would be an example that if the emissions are in Sulfur with a MW of 32. and the model parameters is SO2. We could have the code do this conversion more generally. Rather than hard coding it for all options.

bbakernoaa:
@lwcugb this should be closed now right?

PR #116, opened 2025-02-28
Add Wesely and Zhang dry dep schemes

bbakernoaa:
Add different drydep processes for gas and zhang et al dry deposition 

Note: #111 should be merged first

PR #111, opened 2024-12-20
Add MEGAN2 BVOC emission process

lwcugb:
This pull request adds the MEGAN2.1 BVOC emission scheme based on its implementation in HEMCO.

PR #124, opened 2025-04-24
Add Jacob wet dep process

lwcugb:
Add wet deposition process based on Jacob et al (2001). 

Note: pull request #116 and #111 should be merged first


bbakernoaa:
@lwcugb should we close this now?

PR #126, opened 2025-12-16
Remove javascript code that makes the mermaid plots not work correctly on ReadTheDocs right now

rschwant:
The extra code copying button in the javascripts/extra.js file is making the mermaid plots fail to be created on the ReadTheDocs right now: https://catchem.readthedocs.io/en/latest/

There is already a feature in the material theme of mkdocs called content.code.copy that is already on as specified in the mkdocs.yml file. This one works fine with the mermaid plots since it's all consistent in the material theme and achieves the same functionality.

It looks messy to use both of these options anyway.

So let's take out the javascript code that is not working properly.

### DIFF
--- a/docs/assets/javascripts/extra.js
+++ b/docs/assets/javascripts/extra.js
@@ -26,53 +26,6 @@ document.addEventListener('DOMContentLoaded', function() {
     observer.observe(metric);
   });
 
-  // Copy code blocks functionality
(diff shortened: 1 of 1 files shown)


bbakernoaa (approved):
Took a quick look at the docs from the CI build. It appears the mermaid plots are working.

PR #128, opened 2026-02-13
Add offline surface emission read module and SO4 simple chemistry from GOCART

lwcugb:
A new module for surface offline emission read is tested. A new process for SO4 simple chemical conversion from SO2 is added based on GOCART. Local tests are passed, and further tests on the cluster are needed.


bbakernoaa:
@lwcugb please run pre-commit 

pre-commit run --all-files

lwcugb:
> @lwcugb please run pre-commit
> 
> pre-commit run --all-files

Will do. Still debugging the SO4 chemistry process.

src/process/wetdep/schemes/WetDepScheme_JACOB_Mod.F90
```diff
@@ -417,6 +417,11 @@ subroutine compute_jacob( &
                species_tendencies(k, species_idx) = max(0.0_fp, conc(k)) / dpog(k) * AIRMW / species_mw_g(species_idx) * 1.0e6_fp
             end if
 
+            !!!!!!!!TODO: do not run wetdep on H2O2 since it is read from climatology files for now
```
bbakernoaa, line 420: shouldn't this be configurable in the species.yaml?

lwcugb: This is tricky here because we need H2O2 concentration in the scheme but we do not want to update its concentration through the process. If we turn it off in the species.yaml file,  its concentration will not be passed to the scheme calculation. This is only a temporary solution though since H2O2 should be updated through wetdep if it is not read in from climatology. On that note, I do turn off drydep on SO2 and SO4 in the yaml file so that we do not run drydep twice because GOCART implicitly puts drydep within its chemical process.

bbakernoaa: we need to be careful about this.  It would mean that SO4 for instance uses a different dry deposition than the rest of the model.

bbakernoaa: If the gocart function itself is the problem then we should have our own version of it.  We don't have to implicitly use the gocart2g library for things if it doesn't mesh with our system

lwcugb: That is true. I think we can do something to force the drydep frequency to be zero in the gocart call function so that only the chemical gain/loss rate is used when updating their concentrations.

tests/Configs/Default/CATChem_species.yml
```diff
@@ -39,8 +40,21 @@ h2o2:
...
+  description: Nitrate radical
+  is_gas: true
+  is_photolysis: true
+  is_advected: false
+  mw_g: 62.01
+so4:       #This is to match CMAQ name to use the restart file. so4 is used in GEOS-Chem.
+  name: so4
   description: sulfate
   is_aerosol: true
   is_drydep: true
```
bbakernoaa, line 60: wouldn't this do double deposition if its included in the chemical process?

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
```diff
@@ -0,0 +1,673 @@
...
+!!
+!! Generated on: 2026-02-11T13:30:17.194715
+!! Author: Wei Li
+!! Reference: GOCART2G process library SulfateChemDriver function
+module SO4chemScheme_GOCART_Mod
+
+   use precision_mod, only: fp, rae
+   use SO4chemCommon_Mod, only: SO4chemSchemeGOCARTConfig
+   use error_mod, only: CC_SUCCESS, CC_Error
+   use GOCART2G_Process, only: SulfateUpdateOxidants, SulfateChemDriver, DMSemission
```
bbakernoaa, line 28: we do DMS within this? I think this is ok for now but we really want to separate these processes in the future.

tests/Configs/Default/CATChem_species.yml
lwcugb: I did the test in my local machine and did not push it. As you suggested, we should keep using the same scheme and I will try to turn off drydep in the gocart chemical process or just force it to be zero.

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
lwcugb: Yes. I put it in the same file for simplicity considering ever emission process will be in NEXUS in the future. It is not hard to seperate them if needed though.

bbakernoaa: ok.  just make sure you can turn it off if necessary

tests/Configs/Default/CATChem_species.yml
lwcugb: I have forced the dry deposition frequency to be zero in the latest commit. We can avoid running dry deposition twice in this way and keep using the same scheme for all species.

src/process/so4chem/examples/so4chem_example.F90
```diff
@@ -0,0 +1,406 @@
...
+   call process%finalize(rc)
+   if (rc /= CC_SUCCESS) then
+      write(error_unit, *) 'ERROR: Process finalization failed'
+      stop 1
+   end if
+
+   write(output_unit, '(A)') "Process finalized successfully"
+   write(output_unit, '(A)') "Example completed!"
+
+end program so4chem_example
```
quaz115: @lwcugb @bbakernoaa 

end program so4chem_example appears here, but the file continues with orphaned code through [line 406](https://github.com/ufs-community/CATChem/pull/128/changes#diff-e08eb2f6e581694d2d7a2432199664ea50aab1401ccbec025dcc801a7775bddeR406), including a second end program so4chem_example at line 406. Everything after line 98 (error handling, steps 6–12, contains block, subroutines) is outside the program unit and this file will not compile. Either remove the first end program and merge lines 99–406 into the program body with proper declarations, or delete lines 99–406 entirely.

src/core/ConfigManager_Mod.F90
```diff
@@ -1453,8 +1458,33 @@ subroutine parse_config_data(this, rc)
       call safe_yaml_get_logical(this%yaml_data, 'diagnostics/output/enabled', this%config_data%runtime%DiagEnabled, local_rc)
       if (local_rc /= 0) this%config_data%runtime%DiagEnabled = .false.  ! default value
 
+      call safe_yaml_get_integer(this%yaml_data, 'diagnostics/output/compress_lev', this%config_data%runtime%CompressLev, local_rc)
+      if (local_rc /= 0) this%config_data%runtime%CompressLev = 0  ! default value
+
+      call safe_yaml_get_integer(this%yaml_data, 'diagnostics/output/compress_lev', this%config_data%runtime%CompressLev, local_rc)
+      if (local_rc /= 0) this%config_data%runtime%CompressLev = 0  ! default value
```
quaz115, line 1459: @lwcugb @bbakernoaa 

Lines 1461–1462 and 1464–1465 are identical compress_lev parses? Remove the duplicate at lines 1464–1465.

src/core/ConfigManager_Mod.F90
quaz115: @lwcugb @bbakernoaa 

duplication at lines 1301 and 1304 in config_data_init:


       ! Initialize emission mapping
       call this%emission_mapping%init()

       ! Initialize emission mapping
       call this%emission_mapping%init()

 emission_mapping%init() is called twice with an identical comment.

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
```diff
@@ -0,0 +1,677 @@
...
+      real(fp), pointer, dimension(:,:) :: SU_PSO4aq ! vertical sum of SO4 Prod from aqueous SO2 oxidation [kg/m2/s]
+      real(fp), pointer, dimension(:,:,:) :: SU_emis   ! DMS emissions in kg/m2/s
+      real(fp), pointer, dimension(:,:,:) :: pso2  ! SO2 Prod from DMS oxidation [kg/kg/s]
+      real(fp), pointer, dimension(:,:,:) :: pmsa  ! MSA Prod from DMS oxidation [kg/kg/s]
+      real(fp), pointer, dimension(:,:,:) :: pso4  ! SO4 Prod from all SO2 oxidation [kg/kg/s]
+      real(fp), pointer, dimension(:,:,:) :: pso4g  ! SO4 Prod from gaseous SO2 oxidation [kg/kg/s]
+      real(fp), pointer, dimension(:,:,:) :: pso4aq  ! SO4 Prod from aqueous SO2 oxidation [kg/kg/s]
+      real(fp), dimension(:,:), allocatable :: drydepfrequency
+      real(fp), dimension(:,:), allocatable :: dmso_conc !DMS source concentration in ocean water [nmol/L]
+      !some clock variables for oxidants updates
```
quaz115: @lwcugb @bbakernoaa 
save variables inside a subroutine can make this kernel stateful and not thread-safe. When run_column is called in parallel (e.g., OpenMP over columns), all threads share these save variables and will race on firsttime, nymd_last, nhms_last_recycle, and xh2o2_init. This can produce incorrect results or segfaults in any multi-threaded run.

Suggestion: Move these to members of SO4chemSchemeGOCARTConfig (or a new persistent state type) initialized once during process_init, and pass them into compute_gocart as explicit arguments?

Such as:
```
! In SO4chemSchemeGOCARTConfig or a new SO4chemGOCARTState type:
type :: SO4chemGOCARTState
   logical :: firsttime = .true.
   integer :: nymd_last = -1
   integer :: nhms_last_recycle = -1
   real(fp), allocatable :: xh2o2_init(:,:,:)
```

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
```diff
@@ -0,0 +1,677 @@
...
+   !! \param [INOUT] z0h
+   !! \param [INOUT] u10m
+   !! \param [INOUT] v10m
+   !! \param [INOUT] fraclake
+   !! \param [INOUT] gwettop
+   !! \param [OUT] rc
+   !!
+   !! \ingroup core_modules
+   !!!>
+   pure subroutine PrepMetVarsForGOCART(km,              &
```
quaz115: This subroutine is declared "pure" but contains allocate statements for pointer arguments ([lines ~581–593](https://github.com/ufs-community/CATChem/pull/128/changes#diff-846a585f9f6d0c26f6197fb2b4bc2e1c522dbcbec12ff9436a6e1c70afd7f51aR581)). In standard Fortran, pure procedures cannot have side effects — allocating pointer arguments violates this. Some compilers may accept it, but others will reject it, making the code non-portable.

Suggested change: Remove the pure attribute:

 `  subroutine PrepMetVarsForGOCART(km, ...`

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
```diff
@@ -0,0 +1,677 @@
...
+      !cleanup pointers
+      if (associated(GOCART_TMPU)) deallocate(GOCART_TMPU)
+      if (associated(GOCART_RHOA)) deallocate(GOCART_RHOA)
+      if (associated(GOCART_HGHTE)) deallocate(GOCART_HGHTE)
+      if (associated(GOCART_DELP)) deallocate(GOCART_DELP)
+      if (associated(GOCART_cloud)) deallocate(GOCART_cloud)
+      if (associated(GOCART_PRESS)) deallocate(GOCART_PRESS)
+      if (associated(GOCART_LWI)) deallocate(GOCART_LWI)
+      if (associated(GOCART_USTAR)) deallocate(GOCART_USTAR)
+      if (associated(GOCART_LWI)) deallocate(GOCART_LWI)
```
quaz115: @lwcugb @bbakernoaa GOCART_LWI is deallocated twice. After the first deallocate statement, the pointer becomes undefined (not necessarily disassociated), so the second associated() check may pass and cause a double-free runtime error.

Remove the duplicate line ([line 459](https://github.com/ufs-community/CATChem/pull/128/changes#diff-846a585f9f6d0c26f6197fb2b4bc2e1c522dbcbec12ff9436a6e1c70afd7f51aR459)).

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -535,13 +560,27 @@ subroutine catchem_emis_apply(category, icat, global_scale, config_manager, chem
...
-                  if (emission_flux(i,j,k) > 0.0_fp) then
-                     do k = 1, nz
-                        ! Step 1: Convert to mass mixing ratio change (kg/kg) from emission (kg/m2/s)
-                        ! Step 2: Convert to kg/kg or ppmv using converter calculated above
-                        species_tendency(i,j,k) = emission_flux(i,j,k) * scale_factor *dt * g0 / met_state%DELP(i,j,k) * converter
-                     end do
-                  end if
+                  do k = 1, nz
+                     if (emission_flux(i,j,k) > 0.0_fp) then
+                        if (category_name == 'dms' .and. field_name == 'dms') then
```
quaz115: Unit conversions are buried in hardcoded if/else branches ([Lines 565-576](https://github.com/ufs-community/CATChem/pull/128/changes#diff-b79c71c897864bca5ea80c957840244a0760b81591ace4e7546ed778a511021aR583)) keyed on string category/field names ('dms', 'gmi', 'oh'). This is fragile, especially when we would be adding any new emission category, which requires editing this subroutine. The unit metadata (units field) already exists in the emission YAML config, but is not used here?

Suggestion: Drive the conversion logic from the units field in the emission config? (already parsed into category%fields(ifield)%units), e.g.:

```

select case (trim(category%fields(ifield)%units))
  case ('nmol/l', 'nmol/L')
    species_tendency(i,j,k) = emission_flux(i,j,k) * scale_factor
  case ('#/cm3')
    species_tendency(i,j,k) = emission_flux(i,j,k) * scale_factor &
       / AVO * AIRMW / met_state%AIRDEN(i,j,k) * 1.e3

```

drivers/nuopc/catchem_nuopc_interface.F90
```diff
@@ -1336,6 +1374,211 @@ subroutine write_diagnostic_field(cc_wrap, field_name, data_type, scalar_value,
...
+                  null(), null(), converted_conc, &
+                  trim(description), trim(units_str), filename, rc)
+
+               deallocate(converted_conc)
+            else
+               ! Write gas data directly (already in ppm)
+               call write_diagnostic_field(cc_wrap, field_name, DIAG_REAL_3D, 0.0_fp, &
+                  null(), null(), conc_data, &
+                  trim(description), trim(units_str), filename, rc)
+            end if
```
quaz115, line 1507: @lwcugb @bbakernoaa 

Wouldn't passing null() as actual arguments to write_diagnostic_field for the optional pointer arguments (array_2d_ptr, array_3d_ptr) be risky? Also, a bit non-standard in Fortran when the dummy arguments are optional but not pointers? If the dummies are declared as plain optional arrays, passing null() is illegal Fortran. This may crash at runtime or fail with certain compilers.

Suggestion: Either declare the dummy arguments as pointers, optional in write_diagnostic_field, or omit those arguments using keyword syntax:


```
            call write_diagnostic_field(cc_wrap, field_name, &
               DIAG_REAL_3D, 0.0_fp, &
               array_3d_ptr=converted_conc, &
               description=trim(description), &
               units=trim(units_str), &
               filename=filename, rc=rc)
```

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -1109,7 +1125,18 @@ subroutine catchem_emis_setup_timing(category, clock, frequency, rc)
...
-      category % irec = int( (currTime - startTime) / timeInterval )
+      ! Handle monthly frequency specially due to varying month lengths
+      if (trim(frequency) == "monthly") then
+         ! Calculate actual number of months between start and current time
+         call ESMF_TimeGet(currTime, mm=curr_month, rc=localrc)
+         if (ESMF_LogFoundError(rcToCheck=localrc, msg=ESMF_LOGERR_PASSTHRU, &
+            line=__LINE__,  file=__FILE__,  rcToReturn=rc)) return  ! bail out
+         ! For monthly frequency, set irec to month number (0-based)
+         category % irec = max(0, curr_month -1)
+      else
```
quaz115, line 1148: @lwcugb @bbakernoaa 

For monthly frequency, irec is set to curr_month - 1 here (=0, if 1st month of the current year?). Unless this assumes the input file has exactly 12 records per year, and that the simulation always references the current year's month. Would this work: **If a simulation spans multiple years or the file has multi-year monthly records, this will always read from the same 12-month window and never advance past month 12.**

Suggestion: Rather compute the total number of months elapsed from startTime? For e..g., :

```

      if (trim(frequency) == "monthly") then
         call ESMF_TimeGet(currTime, yy=curr_year, mm=curr_month, rc=localrc)
         call ESMF_TimeGet(startTime, yy=start_year, mm=start_month, rc=localrc)
         category % irec = (curr_year - start_year) * 12 &
            + (curr_month - start_month)
```

src/process/drydep/schemes/DryDepScheme_GOCART_Mod.F90
```diff
@@ -251,18 +251,18 @@ subroutine compute_gocart( &
...
+      if (associated(GOCART_HGHTE)) deallocate(GOCART_HGHTE)
+      if (associated(GOCART_U10)) deallocate(GOCART_U10)
+      if (associated(GOCART_V10)) deallocate(GOCART_V10)
+      if (associated(GOCART_FRACLAKE)) deallocate(GOCART_FRACLAKE)
+      if (associated(GOCART_GWETTOP)) deallocate(GOCART_GWETTOP)
+      if (associated(GOCART_LWI)) deallocate(GOCART_LWI)
+      if (associated(GOCART_USTAR)) deallocate(GOCART_USTAR)
+      if (associated(GOCART_LWI)) deallocate(GOCART_LWI)
+      if (associated(GOCART_HFLUX)) deallocate(GOCART_HFLUX)
+      if (associated(GOCART_Z0H)) deallocate(GOCART_Z0H)
```
quaz115: @lwcugb @bbakernoaa 
 old code used nullify(), the new code uses deallocate(). This is correct only if these pointers were allocated with allocate() in the same scope. If any of these pointers were set via pointer assignment (e.g., ptr => target), deallocate will cause a runtime error. Please confirm that every pointer listed here was allocated (not assigned) in PrepMetVarsForGOCART.

src/core/ChemSpeciesUtils_Mod.F90
```diff
@@ -190,21 +190,18 @@ subroutine create_species_mapping(container, process_species, species_mapping, r
...
       call get_species_indices(container, process_species, species_mapping, rc)
       if (rc /= CC_SUCCESS) return
 
       ! Check for unmapped species and warn
       do i = 1, size(species_mapping)
          if (species_mapping(i) == 0) then
-            write(message, '(A,A,A)') 'Process species "', trim(process_species(i)), &
-               '" not found in chemical mechanism'
-            call CC_Warning(message, rc, 'create_species_mapping')
+            write(*, '(A,A,A)') 'CC_Warning: Process species "', trim(process_species(i)), &
```
quaz115: @lwcugb @bbakernoaa 

Was this change form  CC_Warning and the error manager for structured warning reporting to a raw write(*,*) to stdout, which bypasses the logging infrastructure a conscious choice for any specific reason? Since write(*,*) statements as we know will not be captured in log files during MPI runs (only rank 0 typically sees stdout). The error_mgr pointer that was removed may be still needed?

```

      error_mgr => container%get_error_manager()
      ...
            write(message, '(A,A,A)') 'Process species "', &
               trim(process_species(i)), &
               '" not found in chemical mechanism'
            call CC_Warning(message, rc, 'create_species_mapping')
```

tests/Configs/Default/CATChem_emission.yml
```diff
@@ -18,8 +18,29 @@ anthro:
...
+    long_name: "Nitrate Radical"
+    units: "mol/mol"
+    scale: [1.0]
+    map: ["no3"]
+dms:
+  dms:
+    long_name: "Dimethyl Sulfide"
+    units: "nmol/l"
+    scale: [1.0]
+    map: ["dms_in"]
```
quaz115, line 43: @lwcugb @bbakernoaa 

tests/CATChem_emission.yml vs tests/Configs/Default/CATChem_emission.yml
The DMS emission field is mapped to "dms" in the top-level config but to "dms_in" in the 
 Based on the SO4chem code where dms_in represents the ocean surface DMS concentration (input for emission calculation) and dms is the atmospheric DMS species, the Default config (dms_in) appears correct. Fix and specify the top-level config (the final yaml if there is redundancy at the momet?) to also use dms_in.

src/core/constants.F90
```diff
@@ -44,6 +44,8 @@ module Constants
    !! \brief Constants related to atmospheric composition and properties
    !! \{
    REAL(fp), PARAMETER, PUBLIC :: Cp = 1.0046e+3_fp                !< Specific heat of dry air at constant pressure [J/kg/K]
+   REAL(fp), PARAMETER, PUBLIC :: Cpd = 1.00416e+3_fp              !< Specific heat of dry air at constant pressure [J/kg/K]; value in GOCART
+   REAL(fp), PARAMETER, PUBLIC :: Cpm = 1.00464e+3_fp              !< Specific heat of dry air at constant pressure [J/kg/K]; value in GOCART
```
quaz115, line 48: @lwcugb 

Is this used after this declaration or is this obsolete?

drivers/nuopc/catchem_nuopc_interface.F90
```diff
@@ -1012,6 +1042,7 @@ subroutine transform_catchem_to_field(cc_wrap, field, field_map, rc)
          ! Reverse vertical layers
          do v = 1, nv
             v_cc = cc_wrap%tracer_map%nuopc_to_cc(v)
+            if (.not. chem_state%ChemSpecies(v_cc)%is_advected) cycle !if not advected, go to next cycle
```
quaz115: @lwcugb @bbakernoaa 

The is_advected check is done before v_cc > 0 check, right? 

Here, If v_cc <= 0, accessing chem_state%ChemSpecies(v_cc) is an out-of-bounds array access (index 0 or negative). The v_cc > 0 guard must come first.

Shouldn't this be commented:` !if (.not. chem_state%ChemSpecies(v_cc)%is_advected) cycle line)`  ? Can you clarify this more

drivers/nuopc/catchem_nuopc_interface.F90
```diff
@@ -403,6 +411,11 @@ subroutine catchem_nuopc_run( cc_wrap, dt, current_time, errmsg, rc)
       errmsg = ''
 
       ! Update extemission data first
+      ! Initialize AQMIO component if not done
+      if (.not. ESMF_GridCompIsCreated(cc_wrap%iocomp)) then
+         cc_wrap%iocomp = AQMIO_Create(cc_wrap%grid, rc =rc)
+         if (rc /= CC_SUCCESS) return
+      end if
```
quaz115: @lwcugb @bbakernoaa 

Isn't this initializing the IO component inside the run loop? Which would mean that every timestep pays for this check? Should this be done once during initialization (catchem_nuopc_init), not in the run path?

tests/CATChem_emission.yml
```diff
@@ -18,8 +18,29 @@ anthro:
...
+    long_name: "Nitrate Radical"
+    units: "mol/mol"
+    scale: [1.0]
+    map: ["no3"]
+dms:
+  dms:
+    long_name: "Dimethyl Sulfide"
+    units: "nmol/l"
+    scale: [1.0]
+    map: ["dms"]
```
quaz115: @lwcugb @bbakernoaa 

In tests/CATChem_species.yml, oh, no3, and msa are added **but dms_in is not added**. In tests/Configs/Default/CATChem_species.yml, dms_in is added. The SO4chem process requires dms_in as one of its 8 explicit species, right? Tests using tests/CATChem_species.yml will fail at species lookup. Is this a bug to be fixed?

drivers/nuopc/aqmio.F90
```diff
@@ -3200,6 +3228,28 @@ subroutine AQMIO_VariableCreate(IOLayout, field, unlimited, varId, rc)
...
+      else
+         ncStatus = nf90_def_var_deflate(IOLayout % ncid, lvarId, shuffle=1, deflate=1, deflate_level=0)
+      end if
+      if (ncStatus /= NF90_NOERR) then
+         ! Compression failure is not fatal - continue without compression
+         ! This handles cases where NetCDF4 is not available
+         call ESMF_LogWrite("Warning: Could not enable compression for variable: "//trim(fieldName), &
+            ESMF_LOGMSG_WARNING)
+      end if
+
```
quaz115, line 3252: @lwcugb @bbakernoaa 

Just curious about this:  compression applied unconditionally even when CompressLev=0?

L 3243:  The else branch (when CompressLev is not present) still calls nf90_def_var_deflate with deflate=1 and deflate_level=0. This enables compression machinery even when no compression is wanted; isn't this adding computational overhead? When no compression is desired: nf90_def_var_deflate should either not be called at all, or be called with shuffle=0, deflate=0?

quaz115 (changes requested):
Thanks for the work on this, @lwcugb. My comments are on compile/runtime blockers, a few duplicated logic(s), also MPI thread-safety concerns with shared state in column-parallel calls, hardcoded values that should be config-driven (see .yml file comments), and some general cleanup. Please see the individual comments.

For a PR of this scope (This is quite a significant work in terms of scope of a PR), feel free to add reviewers when opening a PR— feel free to tag anyone relevant besides @bbakernoaa and @rschwant , as well (You can consult @rschwant and @jianheACM on who to tag as a reviewer on relevant PR).

src/core/metstate_mod.F90
```diff
@@ -1214,9 +1214,16 @@ subroutine metstate_derive_field(this, field_name, error_mgr, time_state, rc)
 
       character(len=256) :: thisLoc
       integer :: nx, ny, nz, i, j, k, nlanduse
-      real(fp) :: airden
+      real(fp) :: airden, rh, air_mass
       real(fp) :: avgw ! Water vapor volume mixing ratio [v/v dry air]
       real(fp) :: xh2o ! Water vapor mole fraction [mol (H2O) / mol (moist air)]
+      !some variables used for reevaporation calculations
+      real(fp) :: flux_liq, flux_ice, flux_tot, reevap_liq, reevap_ice,C_evap, C_evap_mixed,RH_term, frac_liq
```
quaz115: @lwcugb @bbakernoaa 

C_evap_mixed is declared but never used. C_evap is computed (temperature-dependent interpolation) but then C_evap_liq and C_evap_ice are used directly in steps 5 and 6 instead of C_evap. The temperature-dependent blending logic is effectively dead code, as currently.

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
```diff
@@ -0,0 +1,677 @@
...
+         !call CC_Error(trim(ErrMsg), RC, thisLoc)
+         write(*,'(A)') trim(ErrMsg)
+         return
+      end if
+
+      !call GOCART sulfate chemistry driver
+      !force dz to be a big value (negative will not work depending on compiler) at the surface to make drydep frequency equal zero.
+      !Five functions need to be customized here if we want to turn it off compleltely.
+      !This is to ensure dry deposition does not run twice for SO2 and SO4
+      GOCART_HGHTE(:,:,num_layers - 1) = GOCART_HGHTE(:,:,num_layers) + 1.0e38_fp
```
quaz115, line 404: @lwcugb @bbakernoaa 

This relies on the GOCART internal convention that dry deposition frequency = dvel / dz for the surface layer. Inflating dz to ~1e38 effectively disables deposition, which avoids double-counting with CATChem's own dry deposition process (is that the intention here?). The intent is documented in the surrounding comments, but consider adding a brief note referencing where in the GOCART library this dz-based calculation occurs, so future maintainers can verify the assumption still holds if the GOCART process library is updated.

bbakernoaa: I really don't like hacking through GOCART2G. It's functional for now and to get us going but I don't think this should be a longterm solution.

drivers/nuopc/catchem_emis_mod.F90
bbakernoaa: I think this is somewhat a short term issue.  We will likely not need this in the future and will likely move to using CDEPS-inline functionality in the future to handle this

drivers/nuopc/aqmio.F90
bbakernoaa: I think this is another short-term issue as when we couple we will likey be able to use the WGC

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
quaz115: @bbakernoaa Thanks for confirming the dz inflation hack (GOCART_HGHTE = 1e38). But at minimum, would suggest adding a comment referencing the specific GOCART library function where dvel/dz is computed so this assumption can be verified when GOCART2G is updated? @lwcugb

drivers/nuopc/aqmio.F90
quaz115: @bbakernoaa @lwcugb Understood — once CATChem moves to the Write Grid Component, the custom AQMIO_VariableCreate compression block goes away. Can we add a TODO on the else branch at line 3243, noting this should be removed when the WGC transition happens, so it doesn't get lost?

drivers/nuopc/catchem_emis_mod.F90
quaz115: @bbakernoaa @lwcugb Thanks for confirming the CDEPS-inline plan for the future, but that means that irec = max(0, curr_month - 1) ignores the year entirely? Any multi-year simulation will cycle through months 0–11 repeatedly and silently read the wrong record once the run crosses a year boundary. 

Would be worth adding a guard of sorts currently that emits an ESMF_LogWrite warning if curr_year /= start_year, so users know this path is untested for multi-year runs until the CDEPS-inline replacement is in place?

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
lwcugb: I did a test with this method and explicitly turned off drydeposition function in GOCART. And they are the same, which means this method works in avoiding running drydep twice. I also considered customizing those GOCART functions in my own scheme file. But that would involve at least five functions from the GOCART library since some local functions are not made public for external calls. This would require more maintenance work once GOCART updates. I can add some notes pointing to the dvel/dz codes in GOCART as suggested.

quaz115: Thanks @lwcugb, good to know you validated it. Adding the reference notes is enough for now in that case.

drivers/nuopc/aqmio.F90
lwcugb: Thanks @quaz115, for noticing that. I think I set deflate_level = 0 and forgot to set the other two to zeros as well.

tests/CATChem_emission.yml
lwcugb: Yes. These two files should be the same, and I mainly used the Default one for testing. Will fix it.

quaz115:
@lwcugb @bbakernoaa

Also some clarifications regarding code workflow (if you can confirm this?): On develop, the old CCPP wrappers (catchem_gocart_wrapper.F90, .meta) have been removed and replaced by [ccpp_catchem_interface.F90](https://github.com/ufs-community/CATChem/blob/develop/drivers/ccpp/ccpp_catchem_interface.F90), which calls catchem_run → ProcessInterface framework — not gocart_chem_mod. However, [gocart_chem_mod.F90](https://github.com/ufs-community/CATChem/blob/develop/src/gocart_chem_mod.F90) still exists in src/ on develop. Is it still compiled/ linked, or is it now dead code? If dead code, should it be removed as part of this PR or a follow-up?

The new path added by this PR ([SO4chemScheme_GOCART_Mod.F90::compute_gocart](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90) → GOCART2G SulfateChemDriver/SulfateUpdateOxidants/ DMSemission) is invoked through the ProcessInterface framework, which is the active CCPP path on develop via ccpp_catchem_interface → catchem_run.

Cross-referencing for some unit conversion and variable indexing and initializing issues to be clarified in the new scheme against the old [gocart_chem_mod.F90](https://github.com/ufs-community/CATChem/blob/develop/src/gocart_chem_mod.F90) for correctness:

1. Oxidant units? — Lines [200–202](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L200-L202) label h2o2_clim and no3_clim as [# cm-3], but lines [284–286](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L284-L286) convert all three with *1.0e-6 (ppm→mol/mol). SulfateUpdateOxidants converts mol/mol→molecules/cm³ internally — so the code is correct but the comments are wrong. The old path handled this differently: [gocart_chem_mod.F90 L117–140](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/gocart_chem_mod.F90#L117-L140) applied diurnal scaling to OH and NO3 before passing to chmdrv_su, while the new path delegates that to SulfateUpdateOxidants. Fix comments to say [ppm] (volume mixing ratio) for all three.
2. DMS_IN surface indexing? — Line [378](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L378): species_conc(1, nDMS_IN) reads layer 1. In the old path this was a 2D surface field. The value is not reversed like lines 372–375, which is correct for a surface quantity. Add a comment clarifying DMS_IN is a 2D surface field stored in layer 1 (bottom) of the 3D array, and that it is intentionally not reversed.


3. H2O2 depletion not carried forward? — Line [345](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L345): xh2o2 = xh2o2_init resets H2O2 each call. SulfateChemDriver depletes xh2o2 internally (equivalent to the old [chem_so2 L565–578](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/gocart_chem_mod.F90#L565-L578)), but the depleted value is never saved back to xh2o2_init. The save attribute (line 226) means each grid cell restarts from undepleted climatology. In the old path, H2O2 was local to each column call so this wasn't an issue. Consider saving post-chemistry xh2o2 back after line 397.

4. Docstring says [mol/mol], differs from actual units? — Line [81](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L81) says species_conc is [mol/mol]. Lines [372–375](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L372-L375) show DMS/SO4 in ug/kg, SO2/MSA in ppm, oxidants in ppm, DMS_IN in nmol/L. The old path used uniform ppm for all four species ([gocart_chem_mod.F90 L105–108](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/gocart_chem_mod.F90#L105-L108)). The forward/reverse conversions are consistent, but the docstring needs per-species unit documentation.

quaz115:
Wesely dry deposition scope — informational note (flagging for future reference) @lwcugb can you confirm this? @bbakernoaa @rschwant @drnimbusrain

The existing [drydep_wesely_mod.F90](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90) (used by the old CCPP wrapper path via [catchem_drydep_wrapper.F90](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/drivers/ccpp/catchem_drydep_wrapper.F90)) contains the Wesely (1989) resistance-model framework. The rc subroutine has active NH3 surface-resistance logic (the [is_nh3 block at ~L1065](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L1065)), but this only executes when p_nh3 > 1 (i.e., NH3 is a registered species), which it isn't under GOCART_SIMPLE. The highnh3 flag is [hardcoded .FALSE.](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L192) since the NH3/SO2 ratio check is [commented out at L203-L207](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L203-L207). All NOy species parameter assignments (hstar, dhr, f0, dvj, ddvel for HNO3, NO2, NO3, N2O5, PAN, HONO, organic nitrates) are also commented out. Only sulfate-cycle species have active deposition velocity logic for [GOCART_SIMPLE at L390-L394](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L390-L394).

**This PR's new ProcessInterface dry deposition path (DryDepScheme_GOCART_Mod.F90) likewise does not activate nitrogen species deposition. This is consistent, but worth noting that if nitrogen chemistry is added in CATChem later, the Wesely scheme's NOy/NHx capabilities would need to be re-enabled (or ported to the new ProcessInterface framework) rather than implemented from scratch.**

bbakernoaa:
> @lwcugb @bbakernoaa
> 
> Also some clarifications regarding code workflow (if you can confirm this?): On develop, the old CCPP wrappers (catchem_gocart_wrapper.F90, .meta) have been removed and replaced by [ccpp_catchem_interface.F90](https://github.com/ufs-community/CATChem/blob/develop/drivers/ccpp/ccpp_catchem_interface.F90), which calls catchem_run → ProcessInterface framework — not gocart_chem_mod. However, [gocart_chem_mod.F90](https://github.com/ufs-community/CATChem/blob/develop/src/gocart_chem_mod.F90) still exists in src/ on develop. Is it still compiled/ linked, or is it now dead code? If dead code, should it be removed as part of this PR or a follow-up?
> 
> The new path added by this PR ([SO4chemScheme_GOCART_Mod.F90::compute_gocart](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90) → GOCART2G SulfateChemDriver/SulfateUpdateOxidants/ DMSemission) is invoked through the ProcessInterface framework, which is the active CCPP path on develop via ccpp_catchem_interface → catchem_run.
> 
> Cross-referencing for some unit conversion and variable indexing and initializing issues to be clarified in the new scheme against the old [gocart_chem_mod.F90](https://github.com/ufs-community/CATChem/blob/develop/src/gocart_chem_mod.F90) for correctness:
> 
> 1. Oxidant units? — Lines [200–202](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L200-L202) label h2o2_clim and no3_clim as [# cm-3], but lines [284–286](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L284-L286) convert all three with *1.0e-6 (ppm→mol/mol). SulfateUpdateOxidants converts mol/mol→molecules/cm³ internally — so the code is correct but the comments are wrong. The old path handled this differently: [gocart_chem_mod.F90 L117–140](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/gocart_chem_mod.F90#L117-L140) applied diurnal scaling to OH and NO3 before passing to chmdrv_su, while the new path delegates that to SulfateUpdateOxidants. Fix comments to say [ppm] (volume mixing ratio) for all three.
> 2. DMS_IN surface indexing? — Line [378](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L378): species_conc(1, nDMS_IN) reads layer 1. In the old path this was a 2D surface field. The value is not reversed like lines 372–375, which is correct for a surface quantity. Add a comment clarifying DMS_IN is a 2D surface field stored in layer 1 (bottom) of the 3D array, and that it is intentionally not reversed.
> 3. H2O2 depletion not carried forward? — Line [345](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L345): xh2o2 = xh2o2_init resets H2O2 each call. SulfateChemDriver depletes xh2o2 internally (equivalent to the old [chem_so2 L565–578](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/gocart_chem_mod.F90#L565-L578)), but the depleted value is never saved back to xh2o2_init. The save attribute (line 226) means each grid cell restarts from undepleted climatology. In the old path, H2O2 was local to each column call so this wasn't an issue. Consider saving post-chemistry xh2o2 back after line 397.
> 4. Docstring says [mol/mol], differs from actual units? — Line [81](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L81) says species_conc is [mol/mol]. Lines [372–375](https://github.com/lwcugb/CATChem/blob/92f2d819b43c233472d78ad41814f714b9d64edf/src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90#L372-L375) show DMS/SO4 in ug/kg, SO2/MSA in ppm, oxidants in ppm, DMS_IN in nmol/L. The old path used uniform ppm for all four species ([gocart_chem_mod.F90 L105–108](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/gocart_chem_mod.F90#L105-L108)). The forward/reverse conversions are consistent, but the docstring needs per-species unit documentation.

v2 is not really coupled with CCPP so its moot.

bbakernoaa:
> Wesely dry deposition scope — informational note (flagging for future reference) @lwcugb can you confirm this? @bbakernoaa @rschwant @drnimbusrain
> 
> The existing [drydep_wesely_mod.F90](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90) (used by the old CCPP wrapper path via [catchem_drydep_wrapper.F90](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/drivers/ccpp/catchem_drydep_wrapper.F90)) contains the Wesely (1989) resistance-model framework. The rc subroutine has active NH3 surface-resistance logic (the [is_nh3 block at ~L1065](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L1065)), but this only executes when p_nh3 > 1 (i.e., NH3 is a registered species), which it isn't under GOCART_SIMPLE. The highnh3 flag is [hardcoded .FALSE.](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L192) since the NH3/SO2 ratio check is [commented out at L203-L207](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L203-L207). All NOy species parameter assignments (hstar, dhr, f0, dvj, ddvel for HNO3, NO2, NO3, N2O5, PAN, HONO, organic nitrates) are also commented out. Only sulfate-cycle species have active deposition velocity logic for [GOCART_SIMPLE at L390-L394](https://github.com/ufs-community/CATChem/blob/2c6fc38992e19d7d03c08b5af7bf06ef016bc61c/src/drydep_wesely_mod.F90#L390-L394).
> 
> **This PR's new ProcessInterface dry deposition path (DryDepScheme_GOCART_Mod.F90) likewise does not activate nitrogen species deposition. This is consistent, but worth noting that if nitrogen chemistry is added in CATChem later, the Wesely scheme's NOy/NHx capabilities would need to be re-enabled (or ported to the new ProcessInterface framework) rather than implemented from scratch.**

Good to note but I think that it can be addressed in a future PR.  We currently do not even have gas phase chemistry in CATChem

src/process/so4chem/examples/so4chem_example.F90
lwcugb: This example file is not being compiled, and I am not sure if it is still necessary to keep it. @bbakernoaa Do you think we can remove it?

src/core/ConfigManager_Mod.F90
lwcugb: I think that is due to an accidental auto-completion when I wrote the code. Deleted the duplicates now.

src/core/ConfigManager_Mod.F90
lwcugb: Save as above. Deleted duplicates.

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
lwcugb: Deleted pure as suggested.

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
lwcugb: Done!

drivers/nuopc/catchem_emis_mod.F90
lwcugb: Good idea and done!

drivers/nuopc/catchem_nuopc_interface.F90
lwcugb: The dummy variables are not optional originally. But I think it is better to use optional variables. Changed the function as suggested.

drivers/nuopc/catchem_emis_mod.F90
lwcugb: Here, the design is for a climatology monthly file reading, so it should work fine for now. I added a new frequency 'yearmonth', which will consider multiple-year monthly inputs. But the limitation here is obvious: the input file has to start from the same year as the simulation start_year.  Again, we need CDEPS to handle this more easily and properly. 

`else if (trim(frequency) == "yearmonth") then`
`    ! Calculate actual number of yearmonths between start and current time`
`    call ESMF_TimeGet(currTime, yy=curr_year, mm=curr_month, rc=localrc)`
`    call ESMF_TimeGet(startTime, yy=start_year, mm=start_month, rc=localrc)`
`    category % irec = max(0, curr_month -1) + (curr_year - start_year) * 12`

src/process/drydep/schemes/DryDepScheme_GOCART_Mod.F90
lwcugb: Thanks for the reminder. It is confirmed.

tests/Configs/Default/CATChem_emission.yml
lwcugb: This tests/CATChem_emission.yml  should be updated.

src/core/constants.F90
lwcugb: These are variables copied from the GOCART library for dry and moist air specific heat. Just want to keep them here for future usage.

drivers/nuopc/catchem_nuopc_interface.F90
lwcugb: Thanks for noticing. Moved it after checking `v_cc > 0` .

drivers/nuopc/catchem_nuopc_interface.F90
lwcugb: Right. Moved it to catchem_nuopc_init now.

src/core/metstate_mod.F90
lwcugb: This should be a typo. C_evap should be used in step 5 and 6, instead of C_evap_liq and C_evap_ice. Fixed now.

src/process/so4chem/examples/so4chem_example.F90
bbakernoaa: Yea I think we can remove the examples. It isn’t needed

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
bbakernoaa: I’ve mentioned this before as well. We should minimize the allocation etc in these routines. If we need to bypass gocart2g library and port the code over so be it.  If it isn’t working for our purpose we don’t have to conform to it.

bbakernoaa:
I’m worried about the future use of gocart2g. It seems to be adding a lot of overhead on us just to leverage.

src/process/so4chem/schemes/SO4chemScheme_GOCART_Mod.F90
lwcugb: This is a good suggestion to move these to a more persistent state with each time step. I updated the process generator to handle this situation when persistent variables are needed in a scheme. This will create a new scheme state within the SO4chemSchemeGOCARTConfig type. Other related functions are also added to the generator.

src/core/ChemSpeciesUtils_Mod.F90
lwcugb: The CC_Warning function gave me some weird outputs in the log file, and I temporarily disabled it. I have fixed the CC_Warning and put it back to use now.

src/process/so4chem/examples/so4chem_example.F90
lwcugb: Removed now.

quaz115:
@rschwant Summary of the PR review as you asked (Also, as you also prefer detailed, line‑level review — please check the linked discussion threads i added for the exact itemized high-level summary of lines and fixes): 

**Overview of the PR review items addressed by** @lwcugb 

Orphaned example program / misplaced end-program — discussion, fix: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951118202
Duplicate compress_lev parse lines — report and duplicate removed: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951138200
Duplicate emission_mapping%init() call — report and duplicate removed: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951145629
save() variables / thread-safety — suggested move to persistent state and generator updated to create scheme state: Suggestion: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951176492 Wei change: https://github.com/ufs-community/CATChem/pull/128#discussion_r2983863498
PrepMetVarsForGOCART pure attribute — suggested and removed: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951189442 Wei confirmation: https://github.com/ufs-community/CATChem/pull/128#discussion_r2976767348
GOCART_LWI double deallocate — reported and duplicate removed: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951199526 Wei reply: https://github.com/ufs-community/CATChem/pull/128#discussion_r2976742654
write_diagnostic_field optional pointer handling — suggested and changed to optional args: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951312195 Wei change: https://github.com/ufs-community/CATChem/pull/128#discussion_r2976999841
Monthly-frequency irec limitation — suggested and "yearmonth" option added: Suggestion: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951331580 Wei change: https://github.com/ufs-community/CATChem/pull/128#discussion_r2977427531
Pointer deallocate vs nullify concern — reported and confirmed/adjusted: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951347164 Wei reply: https://github.com/ufs-community/CATChem/pull/128#discussion_r2977431345
IO component initialization in run loop — flagged and moved to init: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951424797 Wei change: https://github.com/ufs-community/CATChem/pull/128#discussion_r2977527883
C_evap_mixed / C_evap dead/typo — reported and fixed to use C_evap: https://github.com/ufs-community/CATChem/pull/128#discussion_r2955669846 Wei change: https://github.com/ufs-community/CATChem/pull/128#discussion_r2977619854
Compression deflate flags / TODO note — pointed out; flags corrected and TODO added: Issue: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951454214 Wei acknowledgements/changes: https://github.com/ufs-community/CATChem/pull/128#discussion_r2962304105 TODO note: https://github.com/ufs-community/CATChem/pull/128#discussion_r2961805650
H2O2 depletion persistence (xh2o2 reset each call) — reported; persistent scheme state added so depletion can be preserved: Report: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951176492 Wei change: https://github.com/ufs-community/CATChem/pull/128#discussion_r2983863498
CC_Warning / error logging — pointed out and restored/fixed: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951370392 Wei reply: https://github.com/ufs-community/CATChem/pull/128#discussion_r2983933991

**Yet to be addressed:**

Docstring species_conc per-species units mismatch — noted in review but docstring not updated: https://github.com/ufs-community/CATChem/pull/128#issuecomment-4094572194 (references in comment point to lines in SO4chemScheme_GOCART_Mod.F90)
Oxidant units comment incorrect (ppm vs #/cm3) — comment mismatch noted but comments not fixed: https://github.com/ufs-community/CATChem/pull/128#issuecomment-4094572194 (see links in that comment to lines ~200–202 and ~284–286)
Unit-conversion logic hard-coded on category names (suggest drive from YAML units) — suggestion made, no change implemented: https://github.com/ufs-community/CATChem/pull/128#discussion_r2951217454
Wesely dry-deposition NOy/NHx capabilities (porting if/when N chemistry is added) — informational / deferred to future PR: https://github.com/ufs-community/CATChem/pull/128#issuecomment-4099468913
Long-term concern about using GOCART2G (overhead/maintenance) — raised but not resolved here: https://github.com/ufs-community/CATChem/pull/128#issuecomment-4114439342

lwcugb:
@quaz115 @bbakernoaa All the unit conversion and comments are fixed now. I am not sure about the plan on how to proceed with GOCART2G library though. Adapting all the GOCART functions in our own scheme files would require more maintenance work for future updates.

quaz115:
> @quaz115 @bbakernoaa All the unit conversion and comments are fixed now. I am not sure about the plan on how to proceed with GOCART2G library though. Adapting all the GOCART functions in our own scheme files would require more maintenance work for future updates.

Ok, to summarize all items are addressed besides these:

### Actionable (this PR)

| # | Item | Link |
|---|------|------|
| 1 | `pre-commit run --all-files` — confirmed to be ran for for space cleaning and typo corrections. (Minor) | [comment](https://github.com/ufs-community/CATChem/pull/128#issuecomment-3922899659) |

### Deferred (future PRs)

| # | Item | Link |
|---|------|------|
| 1 | GOCART2G overhead / consider porting code | [comment](https://github.com/ufs-community/CATChem/pull/128#issuecomment-4114439342) |
| 2 | CDEPS-inline replacement for emission reader | [comment](https://github.com/ufs-community/CATChem/pull/128#discussion_r2961522977) |
| 3 | WGC transition for AQMIO compression | [comment](https://github.com/ufs-community/CATChem/pull/128#discussion_r2961527696) |
| 4 | Wesely dry-dep NOy/NHx re-enablement when N chemistry is added | [comment](https://github.com/ufs-community/CATChem/pull/128#issuecomment-4099625136) |

lwcugb:
Thanks for the helpful summary. I have run the pre-commit command earlier. It is just for space cleaning and typo corrections.

bbakernoaa:
> > @quaz115 @bbakernoaa All the unit conversion and comments are fixed now. I am not sure about the plan on how to proceed with GOCART2G library though. Adapting all the GOCART functions in our own scheme files would require more maintenance work for future updates.
> 
> Ok, to summarize all items are addressed besides these:
> 
> ### Actionable (this PR)
> #	Item	Link
> 1	`pre-commit run --all-files` — confirmed to be ran for for space cleaning and typo corrections. (Minor)	[comment](https://github.com/ufs-community/CATChem/pull/128#issuecomment-3922899659)
> ### Deferred (future PRs)
> #	Item	Link
> 1	GOCART2G overhead / consider porting code	[comment](https://github.com/ufs-community/CATChem/pull/128#issuecomment-4114439342)
> 2	CDEPS-inline replacement for emission reader	[comment](https://github.com/ufs-community/CATChem/pull/128#discussion_r2961522977)
> 3	WGC transition for AQMIO compression	[comment](https://github.com/ufs-community/CATChem/pull/128#discussion_r2961527696)
> 4	Wesely dry-dep NOy/NHx re-enablement when N chemistry is added	[comment](https://github.com/ufs-community/CATChem/pull/128#issuecomment-4099625136)

I agree with you. The long term support of gocart is seeming more like a hindrance than benefit. Let’s do it this way for now and plan on adapting or implementing them in CATChem itself soon.

bbakernoaa:
@quaz115 is this good to go?

quaz115 (approved):
Approved — good to merge.

All requested fixes have been addressed (unit conversions/comments fixed, pre-commit run completed). The remaining items are deferred for follow-up PRs and should be tracked separately:

Evaluate and reduce long-term dependency on GOCART2G (consider porting key routines into CATChem).
Replace CDEPS-inline emission reader with an integrated CATChem reader.
Transition AQMIO compression to WGC as discussed.
Re-enable/port Wesely NOy/NHx deposition logic if/when gas-phase N chemistry is added.

bbakernoaa (approved):

PR #131, opened 2026-04-08
Update MUSICA submodule to latest version 0.15.0

rschwant:
Update MUSICA submodule to latest version 0.15.0

Checked that all tests still pass after the update.

This is needed to use the latest updates from MICM, so we can pull in AM4.1 chemistry with MICM's latest yaml files formatting update.


rschwant:
I added you both Barry @bbakernoaa and Maggie @mbruckner-work as reviewers, but Barry I think Maggie and I can review this moving forward. But since it's the first time updating a submodule I just wanted to confirm, I am not missing anything.

Not sure why the GCC linux build test above is failing either?

bbakernoaa:
@zmoon can you take a quickly look at the CI build issue

zmoon:
@bbakernoaa it looks like all the recent runs had the spack cached. It changed 9mo ago to need `$SPACK_STACK_DIR` in https://github.com/JCSDA/spack/commit/cf9860864bbebca7f653c8d0ce0b558c772152a4 , which [is set](https://github.com/JCSDA/spack-stack/blob/f499eb7a5cddf7763883cb42031784f2b2f3cd34/setup.sh#L3) when you use `setup.sh` from spack-stack.

For now, could just update that job step to set it to `./spack`. Or switch to an older release ✔️. Or switch to [using spack-stack](https://spack-stack.readthedocs.io/en/1.9.3/CreatingEnvironments.html)?

zmoon:
> Checked that all tests still pass after the update.

Note that the MUSICA integration [isn't currently being tested](https://github.com/ufs-community/CATChem/blame/3b1250005c5e4821d3059fc5442a1db2d99f3902/src/external/CMakeLists.txt#L46-L56) in the build. And the basic API test we had (using MUSICA's MICM Fortran API to get the MICM version) was removed.

You also may wish to remove the notes about MUSICA version in there. Looks like it was already out of date before this change.

rschwant:
I took out the notes for the version number of the MUSICA library as we will likely need to update this frequently and I don't think we want this hardwired here.

@mbruckner-work Maggie is working on testing components of the MUSICA library for compilation and running with her updates to the species.yml format, so that will be in her PR later on once she gets this working. If you like we could close this without merging and she can update the MUSICA library in her PR too, so that it is done all at once.

zmoon (approved):

bbakernoaa (approved):

PR #132, opened 2026-04-09
Feature/ocbc

lwcugb:
This PR is to add GOCART2G carbon species process to CATChem. Two bins of OC and BC are enabled. The major function is to convert from hydrophobic to hydrophilic after reading from emissions. The chemical loss function is included but is turned off by default similar to GOCAT2G's settings. This need to be merged after the SO4 PR #128.


quaz115:
@lwcugb  @bbakernoaa  Some high-level comments to start with:

1. Document a bit more the dependency on SO4 PR #128 and explain where sulfate coating/coupling is implemented and why merge order matters?
2. Please list the exact new tracer variable names added by this PR and the corresponding namelist switches. That will help reviewers and users map inputs/outputs precisely.
3. Confirm the new tracers are included in restart and history I/O. Without restart support reproducibility and restarts will break.
4. Provide a short regression or unit check demonstrating total-carbon mass conservation when chemical-loss is disabled.
5. State the intended process ordering explicitly in the code and PR: aging must run after emissions are applied and before deposition/microphysics. 
6. If the chemical-loss option is disabled by default; add a brief rationale in the PR body and relevant code comments explaining that default.
7. Verify units and mass-consistency across emissions mapping, aging, and loss routines; annotate any unit conversions inline.
8. Update namelist and user documentation to list new switches, defaults, expected emission variable names, and any new diagnostics.
9. If new diagnostics or tendencies were added, ensure they are exposed in diagnostics modules and listed in the history field tables.

lwcugb:
@quaz115 Thanks for the comments.
1. The process is not essentially dependent on SO4Chem. It is just because the new code was added before SO4Chem was merged. This new process needs to be merged afterwards to avoid conflicts.
2. The new tracers are added to the [species yaml file](https://github.com/lwcugb/CATChem/blob/feature/ocbc/tests/Configs/Default/CATChem_species.yml). They are OC1, OC2, BC1 and BC2. The new configuration can be found at the end of the [configuration file](https://github.com/lwcugb/CATChem/blob/feature/ocbc/tests/Configs/Default/CATChem_new_config.yml)
3. There is no codes specificly related to restart variables save or read in CATChem. It is handled through host ATM and current AQM workflow in my tests. The model will still run if no variable matching is found in the restart file and the species will start with zeros. 
4. A new function was added in the unit test for a quick mass conversion test now.
5. The process run order is set through the [run_phase section in the configuration file](https://github.com/lwcugb/CATChem/blob/feature/ocbc/tests/Configs/Default/CATChem_new_config.yml#L68)
6. The loss is turned off by defult to match with what GOCART2G is using. I opened an [issue in their repo](https://github.com/GEOS-ESM/GOCART/issues/408) and asked them why. It seems they did some tests on it and decides not to use it by default.  
7. I updated the comments in the `catchem_emis_mod.F90` to make it more clear. Basically, each process follows the unit conversion chain like  
                   `kg/m2/s     ---------->     μg/kg   ------>  kg/kg ------>   μg/kg   -----------> μg/m3`
             `emissions    ×dt⋅g0/ΔP     chemstate         GOCART        chemstate   /AirDen   outputs`

8. Namelists are updated as mensioned above. The documentation is added automatically for each process generated by the process generator. It includes required inputs and outputs as well as process-level diagnostics. Emission is not listed in he documentation for now since it depends on what what inventory and workflow we will be using eventually.
9. This is handled by the process generator automatically too. 

​

src/core/error_mod.F90
```diff
@@ -218,18 +218,18 @@ SUBROUTINE CC_Error( ErrMsg, RC, ThisLoc, Instr )
 
       ! Print error message to log
       Message =  'CATChem ERROR: ' // TRIM( ErrMsg )
-      WRITE(*,'(a)') TRIM(Message)
+      WRITE( 6, '(a)' ) TRIM(Message)
```
bbakernoaa, line 221: Is there a specific reason for the hardwire of the 6 instead of *

bbakernoaa (approved):

quaz115 (approved):
All comments addressed so far, approved

bbakernoaa (approved):

bbakernoaa:
@lwcugb can you resolve the conflicts after merging the SO4 process

src/core/error_mod.F90
lwcugb: It is to be consistent with the previous write line `WRITE( 6, '(a)' ) REPEAT( '=', 79 )` .

lwcugb:
@bbakernoaa Conflicts fixed!

PR #135, opened 2026-04-16
ci: Add matrix builds for CATChem with optional MUSICA support

bbakernoaa:
This pull request introduces support for optionally building the project with the MUSICA chemistry library, updates the minimum required CMake version, and improves the CI workflow to test multiple build configurations. The most important changes include conditional MUSICA integration, C++ standard selection based on build options, and enhancements to the CI matrix for broader coverage.

**Build system enhancements:**

* Updated the minimum required CMake version to 3.21 in `CMakeLists.txt` to support newer features needed for conditional builds.
* Added a new `BUILD_MUSICA` option in `CMakeLists.txt` to allow optional building with the MUSICA chemistry library. The C++ standard is now set to C++20 if MUSICA is enabled, otherwise it defaults to C++11.

**MUSICA integration:**

* Added conditional logic in `src/external/CMakeLists.txt` to configure and include the MUSICA subdirectory only if `BUILD_MUSICA` is enabled, with appropriate build options set for MUSICA and its dependencies.

**Continuous Integration improvements:**

* Updated the GitHub Actions workflow in `.github/workflows/ci.yml` to use a build matrix, testing both the default build and the MUSICA-enabled build, with descriptive job names and appropriate CMake flags.
* Modified the CI build step to pass the correct CMake flags from the matrix, ensuring each configuration is built and tested as intended.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,8 +79,16 @@ jobs:
 
   build-and-test_jcsda-docker:
     runs-on: ubuntu-latest
-    name: Build and test (JCSDA docker image)
+    name: Build and test (JCSDA docker, ${{ matrix.build-label }})
     container: jcsda/docker-gnu-openmpi-dev:1.9
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build-label: CATChem
+            cmake-flags: -DBUILD_NUOPC=ON
+          - build-label: CATChem + MUSICA
+            cmake-flags: -DBUILD_NUOPC=ON -DBUILD_MUSICA=ON
     defaults:
       run:
         shell: bash -l {0}
@@ -93,7 +101,7 @@ jobs:
       - name: Configure and build
         run: |
           source /opt/spack-environment/activate.sh
-          cmake -B $BUILD_DIR -DBUILD_NUOPC=ON
+          cmake -B $BUILD_DIR ${{ matrix.cmake-flags }}
           cmake --build $BUILD_DIR -j2
 
       - name: Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,9 +85,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - build-label: CATChem
+          - build-label: NUOPC
             cmake-flags: -DBUILD_NUOPC=ON
-          - build-label: CATChem + MUSICA
+          - build-label: NUOPC + MUSICA
             cmake-flags: -DBUILD_NUOPC=ON -DBUILD_MUSICA=ON
     defaults:
       run:
--- a/src/external/CMakeLists.txt
+++ b/src/external/CMakeLists.txt
@@ -39,6 +39,18 @@ set_target_properties(
 )
 
 #
+# YAML-CPP — build first so MUSICA/TUV-x can reuse it instead of fetching their own
+#
+
+# Configure yaml-cpp build options
+set(YAML_CPP_BUILD_TESTS OFF CACHE BOOL "Build yaml-cpp tests")
+set(YAML_CPP_BUILD_TOOLS OFF CACHE BOOL "Build yaml-cpp tools")
+set(YAML_CPP_BUILD_CONTRIB OFF CACHE BOOL "Build yaml-cpp contrib")
+
+# Use the local submodule
+add_subdirectory(yaml-cpp)
+
+#
 # MUSICA Fortran API (optional, controlled by BUILD_MUSICA)
 #
 
@@ -58,17 +70,5 @@ if(BUILD_MUSICA)
   add_subdirectory(musica)
 endif()
 
-#
-# YAML-CPP standalone build
-#
-
-# Configure yaml-cpp build options
-set(YAML_CPP_BUILD_TESTS OFF CACHE BOOL "Build yaml-cpp tests")
-set(YAML_CPP_BUILD_TOOLS OFF CACHE BOOL "Build yaml-cpp tools")
-set(YAML_CPP_BUILD_CONTRIB OFF CACHE BOOL "Build yaml-cpp contrib")
-
-# Use the local submodule
-add_subdirectory(yaml-cpp)
-
-# Add our yaml interface (now that yaml-cpp is available independently)
+# Add our yaml interface (yaml-cpp is already available from above)
 add_subdirectory(yaml_interface)
--- a/src/external/CMakeLists.txt
+++ b/src/external/CMakeLists.txt
@@ -45,7 +45,6 @@ set_target_properties(
 if(BUILD_MUSICA)
   # Set MUSICA build options before adding the subdirectory
   set(MICM_ENABLE_TESTS OFF)
-  set(MICM_ENABLE_EXAMPLES OFF)
   set(MUSICA_BUILD_C_CXX_INTERFACE ON)
   set(MUSICA_BUILD_FORTRAN_INTERFACE ON)
   set(MUSICA_ENABLE_INSTALL OFF)
--- a/src/external/CMakeLists.txt
+++ b/src/external/CMakeLists.txt
@@ -39,24 +39,27 @@ set_target_properties(
 )
 
 #
-# MUSICA Fortran API
+# MUSICA Fortran API (optional, controlled by BUILD_MUSICA)
 #
 
-# Set MUSICA build options before adding the subdirectory
-# set(MICM_ENABLE_TESTS OFF)
-# set(MICM_ENABLE_EXAMPLES OFF)
-# set(MUSICA_BUILD_C_CXX_INTERFACE ON)
-# set(MUSICA_BUILD_FORTRAN_INTERFACE ON)
-# set(MUSICA_ENABLE_INSTALL OFF)
-# set(MUSICA_ENABLE_TESTS OFF)
-# set(MUSICA_ENABLE_MICM ON)
-# set(MUSICA_ENABLE_TUVX ON)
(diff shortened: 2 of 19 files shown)


src/external/CMakeLists.txt
```diff
@@ -39,24 +39,27 @@ set_target_properties(
...
+  # Add MUSICA as a subdirectory
+  add_subdirectory(musica)
+endif()
 
 #
-# YAML-CPP standalone build (since we're not using MUSICA)
+# YAML-CPP standalone build
 #
 
 # Configure yaml-cpp build options
```
zmoon: Current error is 
```
CMake Error at src/external/yaml-cpp/CMakeLists.txt:95 (add_library):
  add_library cannot create target "yaml-cpp" because another target with the
  same name already exists.  The existing target is a static library created
  in source directory "/__w/CATChem/CATChem/build/_deps/yaml-cpp-src".  See
  documentation for policy CMP0002 for more details.
```
Can use `NOT BUILD_MUSICA` to guard adding the yaml-cpp.

src/external/CMakeLists.txt
```diff
@@ -39,24 +39,27 @@ set_target_properties(
...
-# set(MUSICA_BUILD_C_CXX_INTERFACE ON)
-# set(MUSICA_BUILD_FORTRAN_INTERFACE ON)
-# set(MUSICA_ENABLE_INSTALL OFF)
-# set(MUSICA_ENABLE_TESTS OFF)
-# set(MUSICA_ENABLE_MICM ON)
-# set(MUSICA_ENABLE_TUVX ON)
+if(BUILD_MUSICA)
+  # Set MUSICA build options before adding the subdirectory
+  set(MICM_ENABLE_TESTS OFF)
+  set(MICM_ENABLE_EXAMPLES OFF)
```
zmoon:
```suggestion
```
Looks like this one doesn't exist anymore. And there are some more options we could add if we want to be fully explicit about what we are not enabling. https://github.com/NCAR/musica/blob/v0.15.0/CMakeLists.txt#L19

zmoon (commented):
Besides the yaml-cpp thing, another issue I had when trying this locally was that `-Werror` was propagating up from MUSICA, causing build to fail since CATChem has many warnings. At the time I got around this by using a Release build instead of the default Debug, but we don't want to have to do that.

src/external/CMakeLists.txt
```diff
@@ -39,33 +39,42 @@ set_target_properties(
...
+  set(MUSICA_BUILD_C_CXX_INTERFACE ON)
+  set(MUSICA_BUILD_FORTRAN_INTERFACE ON)
+  set(MUSICA_ENABLE_INSTALL OFF)
+  set(MUSICA_ENABLE_TESTS OFF)
+  set(MUSICA_ENABLE_MICM ON)
+  set(MUSICA_ENABLE_TUVX ON)
+  set(MUSICA_ENABLE_CARMA OFF)
+
+  # Add MUSICA as a subdirectory
+  add_subdirectory(musica)
```
zmoon, line 71: Need to be linking to get the `-Werror` thing I was mentioning. But I think this should address it for now (using `PRIVATE`, and appending `-Wno-error` to us).

```suggestion
  add_subdirectory(musica)

  target_link_libraries(CATChem_core PRIVATE musica::musica-fortran) 
  target_compile_options(CATChem PRIVATE -Wno-error)
```

src/external/CMakeLists.txt
```diff
@@ -39,33 +39,39 @@ set_target_properties(
 )
 
 #
-# MUSICA Fortran API
+# YAML-CPP — declare via FetchContent so MUSICA/TUV-x reuse it instead of fetching their own
```
zmoon: These and the similar comments needn't and shouldn't mention TUV-x. MICM / MechanismConfiguration use it too (and others, probably more in the future), but those details aren't important.

Also yaml-cpp is usually referred to in all lowercase.

src/external/CMakeLists.txt
```diff
@@ -39,33 +39,39 @@ set_target_properties(
...
+
+if(BUILD_MUSICA)
+  # Set MUSICA build options before adding the subdirectory
+  set(MICM_ENABLE_TESTS OFF)
+  set(MUSICA_BUILD_C_CXX_INTERFACE ON)
+  set(MUSICA_BUILD_FORTRAN_INTERFACE ON)
+  set(MUSICA_ENABLE_INSTALL OFF)
+  set(MUSICA_ENABLE_TESTS OFF)
+  set(MUSICA_ENABLE_MICM ON)
+  set(MUSICA_ENABLE_TUVX ON)
```
zmoon, line 67:
```suggestion
  set(MUSICA_ENABLE_TUVX OFF)
```
might as well, since we're not working on using it yet, just MICM, right?

bbakernoaa: I mean if it builds it wouldn't matter.  It will need to be done. It just builds that library.

bbakernoaa: I'm going to leave it on to prepare for the future.  It doesn't harm anything and will be required.

zmoon: That's fine. The harm in the case is the build taking longer for no (current) reason, but it's negligible.

src/external/CMakeLists.txt
```diff
@@ -39,33 +37,39 @@ set_target_properties(
...
 # Configure yaml-cpp build options
 set(YAML_CPP_BUILD_TESTS OFF CACHE BOOL "Build yaml-cpp tests")
 set(YAML_CPP_BUILD_TOOLS OFF CACHE BOOL "Build yaml-cpp tools")
 set(YAML_CPP_BUILD_CONTRIB OFF CACHE BOOL "Build yaml-cpp contrib")
 
-# Use the local submodule
-add_subdirectory(yaml-cpp)
+# Declare yaml-cpp as already available via FetchContent
+# This prevents MUSICA from fetching their own copy
+FetchContent_Declare(yaml-cpp SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/yaml-cpp)
```
zmoon, line 52: Another option would be that now we could remove the vendored yaml-cpp and fetch it in build like MUSICA does. Unless the goal is to avoid actually `FetchContent`ing things? But if that's a requirement, a similar approach (vendoring or Git submodule) would need to be taken for all of the MUSICA deps.

bbakernoaa: In general for operations we cannot use fetch content. All libraries are required to be installed via GDIT

zmoon: Okay. Even if it is just fetching from GitHub? That's what MUSICA does to fetch MICM and co.

If we need to make them Git submodules instead we can do that now in this PR and like with yaml-cpp mark as available. Or create an issue to remind that it will need to be done at some point for operations compliance.

bbakernoaa: Yes. Any supporting libraries need to be installed by gdit

bbakernoaa: Looking at this yaml-cpp is the only optional library that we really need. Gtest, doxygen, cuda are not required. I think we should move forward.

zmoon: But what about MICM, TUVx, MechanismConfiguration, etc. Those are fetched using `FetchContent` in the MUSICA build process, from GitHub. If that's not allowed, we need to add them as Git submodules to this project and mark as available. Or move on now but open an issue to remind that it needs to be done at some point.

bbakernoaa: We should not rely on their mechanisms. We should be able to have our own and supply the mechanism at runtime. Isn’t this the entire purpose of MICM?

bbakernoaa: We only should need MICM and TUVx

zmoon: I think the MechanismConfiguration is there to facilitate loading mechanisms in their format from YAML/JSON. But the mechanism doesn't have to be one they include (and is loaded at run time).

mbruckner-work: MechanismConfiguration contains the parser tools for reading mechanisms into MICM. As long as the format given there is followed, we can provide our own mechanisms to MICM. 
The chemistry solver process will have a line that loads the provided mechanism file, eg.:
 ` micm => micm_t("../v1/chapman/config.yaml", RosenbrockStandardOrder, error)`

bbakernoaa: This is a later problem.  Let me do some homework to see if we need to ask for a change or if this is acceptable.

zmoon (approved):
Looking good to me.

.github/workflows/ci.yml
```diff
@@ -79,8 +79,16 @@ jobs:
...
-    name: Build and test (JCSDA docker image)
+    name: Build and test (JCSDA docker, ${{ matrix.build-label }})
     container: jcsda/docker-gnu-openmpi-dev:1.9
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build-label: CATChem
+            cmake-flags: -DBUILD_NUOPC=ON
+          - build-label: CATChem + MUSICA
```
zmoon: I know I already approved, but this might make more sense...
```suggestion
          - build-label: NUOPC
            cmake-flags: -DBUILD_NUOPC=ON
          - build-label: NUOPC + MUSICA
```

src/external/CMakeLists.txt
bbakernoaa: I asked NCO about this and they didn’t immediately bulk at it. Given the early timeframe of this let’s kick it down the road.

PR #154, opened 2026-05-15
feat: link MUSICA

zmoon:
Note this builds on #138 ([cmp](https://github.com/mbruckner-work/CATChem/compare/feature/MICM...zmoon:CATChem:try-musica), since been merged) and currently uses https://github.com/NCAR/musica/pull/878 (now [merged](https://github.com/NCAR/musica/commit/beee5e584c030ce6f963f7cefcaaad8d5b7736ba))

Intel oneAPI works on Ursa (tested using [current UWM develop](https://github.com/ufs-community/ufs-weather-model/tree/1e549b4eb30c8c8d52d07857bc7cc62149ade312/modulefiles/)):
```
module use ~/git/ufs-weather-model/modulefiles && module load ufs_ursa.intelllvm && export I_MPI_F90=ifx
cmake -B build-intel -DBUILD_MUSICA=on && cmake --build build-intel -j4
```

Ursa GNU doesn't currently work due to some GFortran 12 issues with CATChem, which I've experimented with addressing in [another branch](https://github.com/zmoon/CATChem/tree/try-musica-gcc-12).
```
module use ~/git/ufs-weather-model/modulefiles && module load ufs_ursa.gnu && export OMPI_FC=gfortran
cmake -B build-gnu -DBUILD_MUSICA=on && cmake --build build-gnu -j4
```

cc: @mbruckner-work

PR #160, opened 2026-06-09
feat: add initial MICM gas phase chemistry solver process

mbruckner-work:
Note this builds on #154.

Adds MICM gas phase chemistry solver to CATChem.  Initial scheme is without photolysis (or "dark earth").
Mechanism configuration included in tests is a simple chemical loss of SO2.


zmoon:
@mbruckner-work I merged in develop and got the other CI jobs passing again. Got the GasChem integration test to work with ctest (`ctest --test-dir build -R Gas --output-on-failure`), but I had to edit the file to get it to provide `micm_config.yaml`.

zmoon:
Well it worked on my laptop but in CI it's  saying
```
Error creating MICM: 
Provided chemical system contains no species.
```

mbruckner-work:
Ran into the same issue from the test with a new build of the code on Hercules.
The test also needs to get passed the reactions.yaml and phases.yaml that are listed in the micm_config.yaml.

I forgot to push an earlier commit that adds the process to the api. How do I add the guard around it for the build without MUSICA?

zmoon:
> Ran into the same issue from the test with a new build of the code on Hercules. The test also needs to get passed the reactions.yaml and phases.yaml that are listed in the micm_config.yaml.

Ah gotcha, I was actually copying those without knowing when I tested it locally, then I changed to a different method for the commit. I can fix.

> I forgot to push an earlier commit that adds the process to the api. How do I add the guard around it for the build without MUSICA?

We need to use preprocessor conditions (`#ifdef MUSICA` or something) and use the `CATCHEM_BUILD_MUSICA` setting to conditionally turn it on via cmake `target_compile_definitions`. But we might want to try to set it up so it's still in the API, but fails with a nice message if MUSICA is going to be used but not available.

Edit: I put in something simple to get it to work for now.

PR #162, opened 2026-06-24
feat: support GFortran 12

zmoon:
* [x] works on *my machine* (GFortran 12.3.0)
* [x] works here in CI (GFortran 12.4.0)
* [x] works on Ursa with UWM GNU modulefiles (GFortran 12.4.0)

note:
* inconsistency with char len for names still exists; probably should introduce parameter variable(s) for it and address more broadly, but that's outside the scope of this PR
  - ending up doing some of this here
* not testing the NUOPC build in CI (gha-esmf only has GCC 13), though we probably want to (e.g. once we're using the new container setups)

xref: #161


src/core/DiagnosticManager_Mod.F90
```diff
@@ -174,13 +174,7 @@ subroutine diagnostic_manager_finalize(this, rc)
```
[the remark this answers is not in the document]
zmoon: `i` is still used but sure we can remove `local_rc`

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         gcc-version: ["13"]
         nuopc: [true, false]
         include:
+          - gcc-version: "12"
+            nuopc: false
```
benkozi: I did some file changes. Will need to update once the semver/container PR merges

src/process/carbchem/ProcessCarbChemInterface_Mod.F90
```diff
@@ -475,12 +475,12 @@ end subroutine run_gocart_scheme_column
...
+   subroutine get_required_met_fields(this, field_names)
       class(ProcessCarbChemInterface), intent(in) :: this
-      character(len=32), allocatable :: field_names(:)
-      character(len=32), allocatable :: scheme_fields(:)
-      character(len=32), allocatable :: process_fields(:)
-      character(len=32), allocatable :: unique_fields(:)
+      character(len=64), allocatable, intent(out) :: field_names(:)
+      character(len=64), allocatable :: scheme_fields(:)
+      character(len=64), allocatable :: process_fields(:)
+      character(len=64), allocatable :: unique_fields(:)
```
benkozi: I wonder if 64 should be constantized. What do you think?

benkozi (changes requested):
Just requesting changes due to the PR dependency

src/process/carbchem/ProcessCarbChemInterface_Mod.F90
zmoon: Yes I think that would be great. And there are other places too; I only changed enough to address the error I was getting. But I would be down to work on that in a separate PR (or here if we think so).

benkozi: It should be a simple add in `src/core/constants.F90`. Can start small in this PR and expand in the future. Your call if you want to do it now or later

zmoon: bit of a rabbit hole, but I gave this an initial go in dd4a9c9, see what you think

256 chosen for consistency with [`NC_MAX_NAME`](https://github.com/Unidata/netcdf-c/blob/a9d00956dd2aab725c7b60a8c047b1958299cc36/include/netcdf.h#L330)

src/core/constants.F90
```diff
@@ -74,6 +74,15 @@ module Constants
    REAL(fp), PARAMETER, PUBLIC :: MWCARB = 12.01e-3_fp             !< Molecular weight of carbon [kg/mol]
    ! \}
 
+   ! \name Character lengths
+   !! \brief Standard maximum lengths for character strings
+   !! \{
+   INTEGER, PARAMETER, PUBLIC :: MAX_LEN_NAME = 256  !< Maximum length for names (fields, config keys, etc.)
+   INTEGER, PARAMETER, PUBLIC :: MAX_LEN_DESC = 1024  !< Maximum length for descriptions
+   INTEGER, PARAMETER, PUBLIC :: MAX_LEN_VERS = 64  !< Maximum length for version strings
+   INTEGER, PARAMETER, PUBLIC :: MAX_LEN_PATH = 512  !< Maximum length for file paths
```
benkozi, line 82: I see what you mean about the rabbit hole. It might be nice to consolidate them, but it looks like only `VERS` could be removed. Otherwise, the changes look good to me

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         gcc-version: ["13"]
```
benkozi: Were these the types of errors you were seeing with GNU?
```
+ OMP_NUM_THREADS=1
+ make -j 8 VERBOSE=1
/scratch3/NCEPDEV/stmp/Benjamin.Koziol/sandbox/git-benkozi/ufs-weather-model/CATChem/src/core/ProcessFactory_Mod.F90:77:52:

   77 |       met_fields = process%get_required_met_fields()
      |                                                    1
internal compiler error: Segmentation fault
Please submit a full bug report,
```
This happened when I tried to compile `develop` with the WM GNU defaults. Just curious

zmoon: Yes that is the first thing I addressed in this branch, working around that ICE by converting it to a subroutine.

src/core/constants.F90
zmoon: If we get rid of the version one, which one are you suggesting to use instead for version strings?

benkozi: Hmmm...256? ESMF uses `ESMF_MAXSTR = 256` for pretty much everything. I can see how that might be too restrictive for desc and path

zmoon: got rid of the version one 23901cb

.github/workflows/ci.yml
zmoon: moved it to the new GNU workflow

src/core/ProcessInterface_Mod.F90
```diff
@@ -32,9 +33,9 @@ module ProcessInterface_Mod
    !!
    type, abstract :: ProcessInterface
       private
-      character(len=64), public :: name = ''         !< Process name
-      character(len=64), public :: version = ''      !< Version string
-      character(len=256), public :: description = '' !< Process description
+      character(len=MAX_LEN_NAME), public :: name = ''         !< Process name
+      character(len=MAX_LEN_NAME), public :: version = ''      !< Version string
+      character(len=MAX_LEN_DESC), public :: description = '' !< Process description
```
benkozi:
```suggestion
      character(len=MAX_LEN_DESC), public :: description = ''  !< Process description
```
Such a super nit. I'm embarrassed for myself

benkozi (approved):
I like these changes 👍

PR #166, opened 2026-07-07
feat(ci): workaround for protected branch push & address semver upgrade issues

benkozi:
## Description

Protected branches allowing only merges/pushes to occur through PRs caused the `release` job to fail. Original testing did not account for this rule. The preferred approach is to install a GitHub App with permissions to modify repository contents and give it an exception in the ruleset. Another option is to tie a personal access token with an exception given to the holder or create a dedicated bot account. The PAT approach is less stable in the long-run as it requires someone always having access to said account and is vulnerable to accidental pushes/mods.

Also included:
- Fixes for an upgrade to the `python-semantic-release` version update
- Fixes a race condition where `docker-release` could be canceled by the tag commit CI

Work remaining:
- [x] Install app on `ufs-community/CATChem`. Someone in EPIC CM can help with this. If that installation falls through, the PAT is fallback.
- [x] Add secrets to the CATChem repository for the app's ID and private key.

## Type of change

- Bug fix (fixes something broken)

## Change characteristics
- Is this a breaking change (a change in existing functionality)? NO
- Does this change require a documentation update? NO

## How has this been tested?

Ran a test version of the CI recreating the branch ruleset.

## Checklist
- [x] Any dependent changes have been merged and published
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [x] I have made corresponding changes to the documentation if necessary

Resolves https://github.com/ufs-community/CATChem/issues/165


benkozi:
Thank you @gspetro-NOAA for helping install the required app!

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # CHANGELOG
 
+<!-- version list -->
```
zmoon, line 3: is this necessary?

pyproject.toml
```diff
@@ -50,6 +50,15 @@ version_toml = [
 version_variables = []
 upload_to_pypi = false
 build_command = ""
+allow_zero_version = true
+commit_message = "{version}\n\nAutomatically generated by python-semantic-release [ci skip]"
+vcs_release = true
+
+[tool.semantic_release.changelog]
+exclude_commit_patterns = []
```
zmoon: is this necessary? seems to be the default. although it also [seems](https://python-semantic-release.readthedocs.io/en/latest/configuration/configuration.html#exclude-commit-patterns) like they encourage you to exclude the old message format (without the ` [ci skip]`). thoughts?

pyproject.toml
```diff
@@ -50,6 +50,15 @@ version_toml = [
...
+
+[tool.semantic_release.changelog.default_templates]
+changelog_file = "CHANGELOG.md"
```
zmoon: also seems to be the [default](https://python-semantic-release.readthedocs.io/en/latest/configuration/configuration.html#config-changelog-default-templates-changelog-file)

zmoon (approved):
looks fine to me

CHANGELOG.md
benkozi: Yes, in fact that's why the additional template call was added to see if it would trigger the change log update. Then I found this version string thing: https://python-semantic-release.readthedocs.io/en/latest/concepts/changelog_templates.html#update-mode

pyproject.toml
benkozi: Hmmm...I don't remember seeing those commits even with `ci skip` appearing during my testing. I could be mistaken though. I'll test again. You're right that it can be removed otherwise.

pyproject.toml
benkozi: Yup, will do. See above comment 👍

rschwant (approved):
I like this direction! Thanks for the updates.

CHANGELOG.md
zmoon: interesting..

pyproject.toml
benkozi: Looking at the p-s-r source I think you're right. Good catch. I updated.

pyproject.toml
benkozi: Removed!

benkozi:
Thanks for the reviews! Hopefully no more follow-ups 🤞

PR #169, opened 2026-07-20
feat: Add standalone driver for offline simulation

lwcugb:
This is to add a CATChem app and driver to the standalone folder so that it can run offline, driven by reanalysis meteorological fields. 

MET fields are still read in through the emission module together with other input files. Additional configuration files were also created in the default folder. An initial test driven by MERRA2 and GFS was working fine.


cmake/generate_metstate_macros.py
```diff
@@ -62,6 +62,11 @@ def parse_metstate_type(filename):
         elif in_type and 'end type' in line.lower():
```
benkozi, line 62: @lwcugb - Would you like this in main for WM integration? If not, we may want to merge to develop after in case changes are required during the WM PR review.

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -259,22 +268,352 @@ subroutine catchem_emis_update(ext_emis_data, current_time, state_manager, IO, g
...
+      !     primaries and derive_field self-resolves the PS -> PEDGE -> PMID ->
+      !     {DELP, AIRDEN} chain from whichever category supplied them.
+      ! MetState's registry is reset each timestep, so this re-derives every step
+      ! from the current PS/T in the offline case (no staleness).
+      if (.not. met_state%is_field_set('DELP')) then
+         call met_state%derive_field('DELP', error_manager, time_state, localrc)
+         if (localrc /= CC_SUCCESS) then
+            call ESMF_LogWrite(trim(pName)//': could not derive DELP (surface pressure '// &
+               'PS may be unavailable); kg/m2/s emission unit conversion may be invalid', &
+               ESMF_LOGMSG_WARNING, rc=localrc)
```
benkozi: Curious why this is not an error?

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -259,22 +268,352 @@ subroutine catchem_emis_update(ext_emis_data, current_time, state_manager, IO, g
...
+   logical function emis_category_is_met(config_manager, icat) result(is_met)
+      type(ConfigManagerType), pointer, intent(in) :: config_manager
+      integer, intent(in) :: icat
+      integer :: ifield, ispec, nmap
+      character(len=64) :: tgt
+
+      is_met = .false.
+      if (.not. associated(config_manager)) return
+      if (.not. config_manager%config_data%emission_mapping%is_loaded) return
+      if (icat < 1 .or. icat > config_manager%config_data%emission_mapping%n_categories) return
```
benkozi, line 360: Why the early returns? MPI-related?

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -259,22 +268,352 @@ subroutine catchem_emis_update(ext_emis_data, current_time, state_manager, IO, g
...
+
+      associate (cat => config_manager%config_data%emission_mapping%categories(icat))
```
benkozi, line 362: For my benefit, why is the introspection necessary as opposed to an explicit met field mapping? Seems a little brittle

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -259,22 +268,352 @@ subroutine catchem_emis_update(ext_emis_data, current_time, state_manager, IO, g
...
+         write(msg,'(A,A,A)') trim(pName), &
+            ': pressure interpolation unavailable, used level copy for ', &
+            trim(category%category_name)//'/'//trim(field%field_name)
+         call ESMF_LogWrite(msg, ESMF_LOGMSG_WARNING, rc=localrc)
+      end if
+
+   end subroutine catchem_emis_vinterp_field
+
+   !> \brief Lowercase an in-place string (ASCII)
+   subroutine to_lower_str(s)
```
benkozi, line 613: Could be used in tandem with some string comparisons (`time', `Time`)

drivers/nuopc/catchem_nuopc_cap.F90
```diff
@@ -58,6 +58,8 @@ module cc_nuopc
 
    public :: SetServices
    public :: SetVM !for GCAFS only
+   public :: cc_nuopc_set_config_file       !< Set CATChem science config (YAML) path
```
benkozi, line 61: Will your feature be affected by Barry's C++ CATChem refactor?

drivers/nuopc/catchem_nuopc_interface.F90
```diff
@@ -687,40 +697,170 @@ subroutine transform_nuopc_to_catchem(cc_wrap, importState, currTime, rc)
...
+   !! \param[out]   any_provided   .true. if at least one required field is met-mapped
+   subroutine emission_provided_met_mask(cc_wrap, mask, any_provided)
+
+      type(cc_wrap_type), intent(inout) :: cc_wrap
+      logical, intent(out) :: mask(:)
+      logical, intent(out) :: any_provided
+
+      type(StateManagerType), pointer :: state_mgr
+      type(ConfigManagerType), pointer :: config_mgr
+      character(len=64) :: tgt
```
benkozi: Could use string len constant here

drivers/standalone/catchem_app.F90
```diff
@@ -0,0 +1,131 @@
...
+
+   implicit none
+
+   character(len=*), parameter :: u_FILE_u = __FILE__
+
+   integer             :: rc, urc
+   type(ESMF_VM)       :: vm
+   type(ESMF_GridComp) :: driver
+   integer             :: localPet
+   character(len=512)  :: driver_cfg_file
```
benkozi: String constant

drivers/standalone/catchem_driver.F90
```diff
@@ -0,0 +1,361 @@
...
+   implicit none
+   private
+
+   public :: SetServices
+   public :: set_driver_config_file
+
+   character(len=*), parameter :: u_FILE_u = __FILE__
+
+   !> Path to the driver ESMF configure file (set by the main application)
+   character(len=512), save :: g_driver_cfg_file = "catchem_standalone.configure"
```
benkozi: String constant police (last time)

drivers/standalone/catchem_standalone_grid_mod.F90
```diff
@@ -0,0 +1,408 @@
...
+         call create_column_grid(cfg, grid, rc)
+       case (GRID_MODE_GRIDDED)
+         call create_gridded_grid(cfg, grid, rc)
+       case default
+         call ESMF_LogSetError(ESMF_RC_ARG_BAD, &
+            msg="Unknown grid mode '"//trim(cfg%mode)//"' (expected 'column' or 'gridded')", &
+            line=__LINE__, file=__FILE__, rcToReturn=rc)
+         return
+      end select
+      if (rc /= ESMF_SUCCESS) return
```
benkozi: Log? Error?

tests/test_MetState.f90
```diff
@@ -170,11 +170,22 @@ program test_MetState
```
benkozi, line 170: There's a lot of new code with no automated testing. What would you recommend? Manual testing is challenging if we need to make changes in the future

benkozi (changes requested):
A very useful feature for research and testing CATChem.

cmake/generate_metstate_macros.py
benkozi, line 59: Do you plan to add documentation on how to run standalone?

drivers/nuopc/catchem_emis_mod.F90
lwcugb: This is not MPI-related. These are just defensive guards to avoid null pointer reference or an out-of-bounds array access on `categories(icat)`.

drivers/nuopc/catchem_emis_mod.F90
lwcugb: It's only used to decide ordering: met-providing categories (all mappings prefixed MET_) must be applied first so PS/T/etc. exist before we derive DELP/AIRDEN, which the real emission categories then divide by. It's a no-op in coupled runs (met comes from the import); it only matters offline where met is delivered as emission categories.
Rather than re-scan the mappings every timestep, I now compute the classification once at init and cache it on the category (is_met). The hot loop just reads the flag. Same behavior, no per-step string parsing.

drivers/nuopc/catchem_emis_mod.F90
lwcugb: Good suggestion. Did some cleanup with this function for string comparison.

drivers/nuopc/catchem_nuopc_interface.F90
lwcugb: Done

drivers/nuopc/catchem_emis_mod.F90
lwcugb: Change it to error now

drivers/standalone/catchem_standalone_grid_mod.F90
lwcugb: Add an error.

tests/test_MetState.f90
lwcugb: It's fine to use manual testing for the core since we don't change it much once stable. I do not worry too much about it.

cmake/generate_metstate_macros.py
lwcugb: It is ok to hold and put it to develop for now.

drivers/nuopc/catchem_nuopc_cap.F90
lwcugb: It seems minimal for the cap itself, but the interface and core will be affected.

cmake/generate_metstate_macros.py
lwcugb: Good idea, and added to the user guide.

docs/developer-guide/integration/nuopc.md
```diff
@@ -441,8 +441,86 @@ program test_nuopc_compliance
...
+The structure is deliberately extensible. To add another component (e.g. a
+data/forcing component or a mediator):
+
+1. `use` the new component's `SetServices`.
+2. Add a `NUOPC_DriverAddComp` call in `SetModelServices`.
+3. Extend the `runSeq::` block in the driver configure file.
+
 ## Best Practices
 
+
```
benkozi: Nit - remove the extra space :-)

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -259,22 +271,355 @@ subroutine catchem_emis_update(ext_emis_data, current_time, state_manager, IO, g
...
+   subroutine catchem_emis_detect_field_ranks(category, filename, rc)
+      !> \brief Auto-detect each field's 2D/3D rank from the NetCDF file.
+      !!
+      !! Opens the file for metadata inspection only (no data is read) and, for
+      !! every field in the category, sets \c field%is_2d based on the actual
+      !! variable rank after stripping a trailing time/record dimension.  This
+      !! lets a single file/category mix 2D and 3D variables while still being
+      !! read in a single data-open.  If the file or a variable cannot be
+      !! inspected, the field keeps the category-level \c is_2d as a fallback.
+      implicit none
```
benkozi: Minor - can probably remove if this is at the module level? Sometimes fortran confuses me though.

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -259,22 +271,355 @@ subroutine catchem_emis_update(ext_emis_data, current_time, state_manager, IO, g
...
+      rc = CC_SUCCESS
+
+      ! Default every field to the category setting first, so any field we
+      ! cannot inspect below simply inherits the configured is_2d.
+      do ifield = 1, category%n_fields
+         category%fields(ifield)%is_2d = category%is_2d
+      end do
+
+      ncStatus = nf90_open(trim(filename), NF90_NOWRITE, ncid)
+      if (ncStatus /= NF90_NOERR) return  ! keep category defaults
```
benkozi: Minor - seems like some info about why we return here would be useful. Is this an error or just log worthy?

drivers/nuopc/catchem_emis_mod.F90
```diff
@@ -408,18 +764,20 @@ subroutine catchem_emis_read(category, IO, grid, nlev, curr_time, rc)
             timeSlice=category % irec, iofmt=AQMIO_FMT_NETCDF, rc=localrc)
          if (ESMF_LogFoundError(rcToCheck=localrc, msg=ESMF_LOGERR_PASSTHRU, &
             line=__LINE__,  file=__FILE__,  rcToReturn=rc)) then
-            ! Clean up field before returning
-            call ESMF_FieldDestroy(esmf_field, rc=localrc)
+            ! Clean up field and close file before returning
+            call ESMF_FieldDestroy(esmf_field, noGarbage=.true., rc=localrc)
```
benkozi: It's always useful to check that the rc coming back from ESMF is okay especially with allocates/destroys. I'd recommend checking every one even after the logging, though you can kind of get away with those being left alone. I see there checked at the end. However, it's so much nice to have error messages at the point of failure as opposed to going back and bisecting.

drivers/standalone/catchem_app.F90
```diff
@@ -0,0 +1,133 @@
...
+   ! Resolve the driver configure file
+   ! ------------------------------------------------------------------
+   call get_command_argument(1, driver_cfg_file)
+   if (len_trim(driver_cfg_file) == 0) then
+      driver_cfg_file = "catchem_standalone.configure"
+   end if
+   call set_driver_config_file(trim(driver_cfg_file))
+
+   if (localPet == 0) then
+      write(*,'(A,A)') "INFO: [catchem_app] Driver configure file: ", &
```
benkozi: Could use esmf logging here?

src/process/settling/ProcessSettlingInterface_Mod.F90
```diff
@@ -350,6 +350,7 @@ subroutine run_gocart_scheme_column(this, column, rc)
       integer, allocatable :: species_mie_map(:)  ! Mapping from process species to MieData indices
```
benkozi: Can you think of any reason this would change UFS-WM answers? Probably not just checking in case you think we should run it through that regression test

benkozi (changes requested):
Thanks for addressing the previous comments. Minor stuff at this point!

bbakernoaa:
Should we try to make the standalone driver independent of NUOPC/ESMF itself?  This seems like it would be a better use for the community and makes it drastically more portable.

drivers/nuopc/catchem_emis_mod.F90
lwcugb: You are right. Removed now.

drivers/nuopc/catchem_emis_mod.F90
lwcugb: This pass only refines is_2d. If the file can't be opened for inspection, fields keep the category-level is_2d and the real data-open (AQMIO_Open) reports any genuine I/O error via ESMF_LogFoundError, so we return rc=success here. Add some comments above it.

drivers/nuopc/catchem_emis_mod.F90
lwcugb: Good suggestion. I added a new log error check for all the ESMF_FieldDestroy calls.

drivers/standalone/catchem_app.F90
lwcugb: Yes and done

src/process/settling/ProcessSettlingInterface_Mod.F90
lwcugb: I think the main changes are for sea salt, which will be slightly lower. The new edits change the Maring correction only applied to dust species to better match GCAFS settings. I guess we may need to redo the regression test.

docs/developer-guide/integration/nuopc.md
lwcugb: Done

lwcugb:
> Should we try to make the standalone driver independent of NUOPC/ESMF itself? This seems like it would be a better use for the community and makes it drastically more portable.

I agree that it would be more portable for an ESMF-free driver. My concern is that the current IO and offline grid creation are all ESMF-based. If we can solve these two heavy dependencies, a new driver without ESMF would be much easier.

rschwant:
> > Should we try to make the standalone driver independent of NUOPC/ESMF itself? This seems like it would be a better use for the community and makes it drastically more portable.
> 
> I agree that it would be more portable for an ESMF-free driver. My concern is that the current IO and offline grid creation are all ESMF-based. If we can solve these two heavy dependencies, a new driver without ESMF would be much easier.

I agree that eventually the goal should be to make the standalone driver independent of NUOPC/ESMF itself. No problem though if this occurs later on once CECE handles the I/O pieces. We could make an issue to describe this future update if needed to keep track of it.

src/process/settling/ProcessSettlingInterface_Mod.F90
benkozi: @lwcugb - Are you comfortable running the WM regression test? I'm happy to tag-up for a working session or provide the basic commands/approach.

drivers/nuopc/catchem_emis_mod.F90
benkozi: Thanks for the explanation

drivers/nuopc/catchem_emis_mod.F90
benkozi: 👍

bbakernoaa:
> > > Should we try to make the standalone driver independent of NUOPC/ESMF itself? This seems like it would be a better use for the community and makes it drastically more portable.
> > 
> > 
> > I agree that it would be more portable for an ESMF-free driver. My concern is that the current IO and offline grid creation are all ESMF-based. If we can solve these two heavy dependencies, a new driver without ESMF would be much easier.
> 
> I agree that eventually the goal should be to make the standalone driver independent of NUOPC/ESMF itself. No problem though if this occurs later on once CECE handles the I/O pieces. We could make an issue to describe this future update if needed to keep track of it.

I think this is the path forward. With the helm ecosystem we can do this without ESMF.

src/process/settling/ProcessSettlingInterface_Mod.F90
lwcugb: Happy to. If you have any notes or tutorials to share, that would be great. A quick meeting would also be helpful.

benkozi: @lwcugb - Provided everything works, running the WM test is pretty straightforward. I recommend running the test on Ursa. It sounds like baselines will change which will require some digging to confirm the diffs are where we expect. Anyway, let me know how it goes, and we can tag-up to work through the logs.

1. Clone the WM making sure submodules are correct.
```
git clone --recurse-submodules -j 8 --branch ${branch} ${giturl} ${clonedir}
```
2. Run the WM test.
```
cd ${clonedir}/tests
./rt.sh -a ${account} -n "catchem_control intelllvm" -e
```
Documentation for `rt.sh`: https://ufs-weather-model.readthedocs.io/en/develop/BuildingAndRunning.html#the-rt-sh-file

lwcugb: Ok. I'll give it a try and let you know how it goes.

benkozi (approved):
Thanks @lwcugb for working through the suggestions

PR #186, opened 2026-08-17
fix(build): upgrade UFS WM to spack-stack v2.1.x

gspetro-NOAA:
This PR contains updates to CATChem that are required for the WM upgrade to spack-stack v2.1.x (see https://github.com/ufs-community/ufs-weather-model/pull/3174). Currently, this PR includes: 
- Update to CMakeLists.txt to work around a bug in the Gaea C6 oneapi compiler.

### DIFF
--- a/src/external/CMakeLists.txt
+++ b/src/external/CMakeLists.txt
@@ -34,6 +34,26 @@ elseif(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
   endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


benkozi:
@gspetro-NOAA - We're leaking changes into `main` that are not quite ready. I think the best approach is to close this PR and re-open using `main` as the base branch. The fix here is a "hotfix" and should not incorporate anything from `develop`. Is that okay? So branch from `main` then PR to `main`.

Thanks @zmoon for the RCA.

